PR #110, opened 2026-04-01
fix: rewrite worker guidelines to explain whys and stay repo-agnostic (#100)

### LINKED ISSUE #100: bug: agents respond in Slack but don't do the actual work
## Problem

Agents receive tasks via `pinet_message`, respond in Slack saying they'll do it, but then don't actually execute the work (no commits pushed, no code changes). They appear to be going idle after posting their acknowledgment.

## Observed behavior

- Broker delegates task with clear instructions (branch, files, what to fix)
- Agent replies in Slack: "On it" / "Starting now" / posts a plan
- No commits appear on the branch
- Agent shows as "idle" in `pinet_agents`
- Work doesn't get done unless broker follows up repeatedly

## Possible causes

1. **Agent loop ends after Slack reply** — the agent treats the `pinet_message` response as completing its task, then goes idle waiting for the next prompt
2. **No follow-through mechanism** — there's nothing keeping the agent in a work loop after acknowledging
3. **Worker guidelines too weak** — prompt says "report progress" but doesn't say "keep working until done"
4. **Inbox drain timing** — agent drains inbox (pinet_message), responds, then has nothing left to process

## Fix ideas

1. **Stronger worker guidelines** — "When you receive a task, work on it until completion. Don't just acknowledge — do the work, push the code, then report done."
2. **Ralph loop nudging** — the ralph loop should detect idle agents with assigned work and nudge them (this is being built in #91)
3. **Task tracking in broker DB** — track assigned tasks, not just thread ownership. If an agent is assigned a task and goes idle without completing it, escalate.
4. **Persistent task queue** — tasks survive across agent prompt cycles

## Related

- #91 (ralph loop — will help with nudging)
- #71 (Pinet delegation)

gugu91:
Closes #100.

Rewrites `buildWorkerPromptGuidelines()` in `slack-bridge/helpers.ts`:

- Explains the **why** behind each guideline (sender needs closure, blocked work should be visible, etc.)
- Keeps the workflow generic for Pinet: ACK → work → ask if blocked → report
- Removes repo-specific instructions (worktrees, validation commands, PR flow) — those come from the local repo prompt/AGENTS.md
- Preserves the `pinet_message` vs Slack thread routing rule

Validation:
- `pnpm lint` ✅
- `pnpm typecheck` ✅
- `pnpm test` ✅ (387 passed)

### DIFF
--- a/slack-bridge/index.ts
+++ b/slack-bridge/index.ts
@@ -806,11 +806,7 @@ export default function (pi: ExtensionAPI) {
     promptSnippet: "Check for new incoming Slack messages.",
     promptGuidelines: [
       "You are connected to Slack via the slack-bridge extension.",
-      "Use the current Slack identity from the system prompt when replying in Slack threads.",
-      "New Slack messages are queued — call `slack_inbox` periodically (e.g. between tasks or when you see the badge count increase) to check for pending messages.",
-      "Reply to each message with `slack_send`, passing the correct `thread_ts`.",
-      "Follow the current system prompt exactly for first-message vs follow-up identity formatting.",
-      "Always use this name and emoji — do not invent a new one.",
+      "When you receive messages: ACK briefly, do the work, report blockers immediately, report the outcome when done.",
       ...(securityPrompt
         ? [
             "Security guardrails are active for Slack-triggered actions. Check the security prompt in each message for restrictions.",
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -29,6 +29,14 @@ pnpm format        # Prettier + Stylua
 Turbo caches lint/typecheck/test results per-package. Unchanged packages
 are skipped on re-runs.
 
+## Worktree rules
+
+- NEVER work directly on `main` or checkout feature branches in the main repo directory.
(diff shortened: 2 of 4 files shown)


slack-bridge/index.ts
```diff
@@ -1443,7 +1447,8 @@ export default function (pi: ExtensionAPI) {
     name: "pinet_message",
     label: "Pinet Message",
     description: "Send a message to another connected Pinet agent.",
-    promptSnippet: "Send a message to another connected Pinet agent.",
+    promptSnippet:
+      "Send a message to another connected Pinet agent. When you receive a task: ACK briefly, do the work, report blockers immediately, report the outcome when done. Always reply where the task came from.",
```
will-jack-and-jilll:
```suggestion
      "Send a message to another connected Pinet agent.",
```

gugu91:
```suggestion
      "Send a message to another connected Pinet agent. When you send a task, sepcify the desired workflow, ideally something like `ack/work/ask/report`: ACK briefly, do the work, report blockers or questions immediately, report the outcome when done. Always reply where the task came from.",
```

PR #112, opened 2026-04-01
fix: allow Ralph loop follow-up repeats after cooldown

gugu91:
## Summary

- let persistent Ralph anomalies re-notify after cooldown instead of suppressing the same signal forever
- keep non-stacking and busy-state suppression for broker follow-ups
- preserve cooldown across transient clean cycles so flapping anomalies do not re-notify immediately

## Validation

- pnpm lint
- pnpm typecheck
- pnpm test

### DIFF
--- a/slack-bridge/helpers.test.ts
+++ b/slack-bridge/helpers.test.ts
@@ -851,6 +851,26 @@ describe("shouldDeliverRalphLoopFollowUp", () => {
       }),
     ).toBe(false);
   });
+
+  it("keeps cooldown active across a transient clean cycle", () => {
+    const deliveredAt = 10_000;
+
+    expect(
+      shouldDeliverRalphLoopFollowUp({
+        signature: "",
+        lastDeliveredAt: deliveredAt,
+        now: deliveredAt + 15_000,
+      }),
+    ).toBe(false);
+
+    expect(
+      shouldDeliverRalphLoopFollowUp({
+        signature: "ghost agents detected: ghost-1",
+        lastDeliveredAt: deliveredAt,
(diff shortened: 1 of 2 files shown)


will-jack-and-jilll:

## Verdict
- **Approve**
- This fixes the Ralph follow-up regression cleanly: repeat alerts are allowed after cooldown, pending follow-ups still do not stack, and cooldown now survives transient clean cycles so flapping anomalies do not re-notify immediately.

## Review Scope
- PR reviewed: `#112 https://github.com/gugu91/extensions/pull/112`
- Files examined: `slack-bridge/helpers.ts`, `slack-bridge/index.ts`, `slack-bridge/helpers.test.ts`
- Commands run: `gh pr diff 112`, `pnpm --dir slack-bridge test -- --runInBand`, `pnpm --dir slack-bridge typecheck`
- PiComms: `pr-112-review` comment posted
- GitHub: `comment posted`

## Findings

### Critical (must fix)
- None.

### Warnings (should fix)
- None.

### Suggestions (consider)
- None.

## Checklist
- Correctness: pass — same-signature anomalies can re-notify after cooldown, pending still blocks stacking, and clean-cycle flaps no longer reset cooldown.
- Architecture & maintainability: pass — removing the extra follow-up signature state makes the delivery logic simpler.
- Security & reliability: pass — behavior is safer against noisy flap loops without changing external trust boundaries.
- Tests & verification: pass — targeted tests were updated and slack-bridge test + typecheck passed.

## Summary
- Fixes the previous forever-suppressed repeat-alert behavior.
- Preserves non-stacking follow-up behavior.
- Preserves cooldown across transient recovery/flap cycles.

PR #151, opened 2026-04-02
fix: cap Slack API retry at 3 attempts to prevent infinite recursion (#124)

### LINKED ISSUE #124: bug: top-level slack() has no retry limit — infinite recursion on sustained 429
## Problem

The `slack()` function in `slack-bridge/index.ts` (line ~65) retries on HTTP 429 with **no maximum retry count**:

```typescript
async function slack(method, token, body) {
  // ...
  if (res.status === 429) {
    const wait = Number(res.headers.get('retry-after') ?? '3');
    await new Promise(r => setTimeout(r, wait * 1000));
    return slack(method, token, body);  // recursive, no limit\!
  }
  // ...
}
```

Under sustained rate limiting, this will hit a stack overflow. The adapter's `callSlack()` in `adapters/slack.ts` correctly caps at `SLACK_MAX_RETRIES = 3`, but the main extension's `slack()` doesn't.

## Where

- `slack-bridge/index.ts` line ~65 — `slack()` function
- Compare with `slack-bridge/broker/adapters/slack.ts` — `callSlack()` has proper retry cap

## Suggested Fix

Add a `_retryCount` parameter and cap at 3 retries, matching `callSlack()`. Or better, deduplicate: delete the top-level `slack()` and use `callSlack()` from the adapter module.

## Severity

**High** — production risk under load. Stack overflow crashes the process.

gugu91:
## Problem

The top-level `slack()` wrapper in `index.ts` retried 429 responses recursively with **no maximum retry count**. Under sustained rate limiting, this would cause a stack overflow.

The adapter's `callSlack()` in `adapters/slack.ts` already caps at `SLACK_MAX_RETRIES = 3`, but the main extension's `slack()` didn't.

## Fix

Added a `_retryCount` parameter and `SLACK_MAX_RETRIES = 3` constant. After 3 retries, throws an error instead of recursing.

Matches the existing behavior in `callSlack()`.

## Testing

- `pnpm lint` ✅
- `pnpm typecheck` ✅
- `pnpm test` ✅ (394 tests pass)

## Files changed

- `slack-bridge/index.ts` — 7 insertions, 1 deletion

Closes #124

### DIFF
--- a/slack-bridge/index.ts
+++ b/slack-bridge/index.ts
@@ -67,18 +67,24 @@ interface SlackResult {
   [k: string]: unknown;
 }
 
+const SLACK_MAX_RETRIES = 3;
+
 async function slack(
   method: string,
   token: string,
   body?: Record<string, unknown>,
+  _retryCount = 0,
 ): Promise<SlackResult> {
   const { url, init } = buildSlackRequest(method, token, body);
   const res = await fetch(url, init);
 
   if (res.status === 429) {
+    if (_retryCount >= SLACK_MAX_RETRIES) {
+      throw new Error(`Slack ${method}: rate limited after ${SLACK_MAX_RETRIES} retries`);
+    }
     const wait = Number(res.headers.get("retry-after") ?? "3");
     await new Promise((r) => setTimeout(r, wait * 1000));
-    return slack(method, token, body);
+    return slack(method, token, body, _retryCount + 1);
   }
 
   const data = (await res.json()) as SlackResult;


gugu91 (commented):
*Will's reviewer agent Retry Ferret*

## Verdict: ✅ Approve

Clean, minimal fix that caps the top-level `slack()` function at 3 retries on 429s, preventing infinite recursion under sustained rate limiting. Mirrors the existing `callSlack()` pattern exactly.

### Key findings
- **No critical or warning issues**
- **Correctness confirmed**: `_retryCount` defaults to 0, increments correctly, throws at ≥ 3. All 20+ callers pass only 3 args — no accidental retry-count injection.
- **Consistency**: Implementation is line-for-line identical to `callSlack()` in `broker/adapters/slack.ts`.
- **Verification**: 394 tests pass, lint clean, typecheck clean.

### Suggestions (non-blocking)
1. **DRY**: `slack()` and `callSlack()` are now structurally identical. Consider extracting a shared helper to avoid future divergence.
2. **Tests**: No unit test for the retry cap path (pre-existing gap). Worth a follow-up if retry logic evolves.

PR #153, opened 2026-04-02
fix: add hard broker guardrails to prevent coding (#107)

### LINKED ISSUE #107: fix: broker agent keeps coding instead of delegating — needs hard guardrails
## Problem

The broker agent (LLM) repeatedly picks up coding tasks itself despite prompt guidelines saying not to. Pattern:

1. User reports a bug
2. Broker investigates code (fine)
3. Broker starts editing files, running tests, pushing commits (NOT fine)
4. User has to interrupt: "buddy"

This happened with #101 (unrouted messages fix — broker coded it directly on main) and again with #106 (started editing schema.ts before catching itself).

## Why it happens

1. **Prompt-only enforcement is weak** — the LLM sees the code, knows the fix, and the path of least resistance is to just do it
2. **No tool restrictions** — broker has full access to `edit`, `write`, `git commit`, `git push`
3. **Urgency bias** — when something is "a big bug right now", the broker skips delegation

## Fix ideas

1. **Restrict broker tools** — broker should NOT have access to `edit` or `write` on source files. Read-only + `pinet_message` + `gh` commands only.
2. **Stronger prompt guardrails** — add explicit "NEVER edit .ts files" rule, with examples of what to do instead
3. **Broker-specific tool allowlist** — `buildBrokerPromptGuidelines()` could inject a tool policy
4. **Self-check hook** — before_agent_start checks if broker is about to edit code and blocks it

## Related

- #100 (lazy workers — if workers actually worked, broker wouldn't feel the need to code)
- #104 (ralph loop followUp — better maintenance loop reduces manual broker intervention)

gugu91:
## Problem

The broker agent keeps picking up coding tasks instead of delegating to workers via Pinet (Issue #107, P0). When the broker codes, Slack messages stop flowing, dead agents don't get reaped, and the whole multi-agent mesh stalls.

### 1. System prompt guardrails (`helpers.ts`)
- **HARD RULE** block: `NEVER WRITE CODE` with mesh-stall justification
- **FORBIDDEN** actions list: Agent tool, edit, write, bash for code changes
- **ALLOWED** actions list: route, delegate, review, inspect
- **Refusal template**: 'I'm the broker — I coordinate, not code.'
- **No fallback**: 'NEVER do the work yourself as a fallback'

### 2. Tool-level guardrails (`guardrails.ts`)
- `BROKER_FORBIDDEN_TOOLS` set blocks the `Agent` tool
- `isBrokerForbiddenTool()` check function
- `buildBrokerToolGuardrailsPrompt()` for system prompt injection

### 3. Runtime enforcement (`index.ts`)
- `tool_call` event handler returns `{ block: true }` for Agent tool when broker role is active — hard block, can't be bypassed by prompt
- Broker tool guardrails prompt injected via `before_agent_start`

## Tests
- 405 tests passing (lint ✅ typecheck ✅ test ✅)
- New tests for `isBrokerForbiddenTool`, `BROKER_FORBIDDEN_TOOLS`, `buildBrokerToolGuardrailsPrompt`
- Updated tests for strengthened `buildBrokerPromptGuidelines`

Closes #107

### DIFF
--- a/slack-bridge/guardrails.test.ts
+++ b/slack-bridge/guardrails.test.ts
@@ -6,6 +6,9 @@ import {
   buildSecurityPrompt,
   isConfirmationApproval,
   isConfirmationRejection,
+  isBrokerForbiddenTool,
+  buildBrokerToolGuardrailsPrompt,
+  BROKER_FORBIDDEN_TOOLS,
   READ_ONLY_TOOLS,
   WRITE_TOOLS,
   type SecurityGuardrails,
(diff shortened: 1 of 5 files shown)


gugu91:
*Will's reviewer agent Turbo Falcon*

## Verdict: ✅ Approve

Three-layer defense is architecturally sound. The runtime `tool_call` block (layer 3) is the real enforcement — can't be prompt-jailbroken. Prompt rules (layers 1-2) are well-written and significantly stronger than before. All tests pass (3222), lint ✅, typecheck ✅.

### ⚠️ Warnings (should fix)

**1. ALLOWED section contradicts Agent tool block** (`helpers.ts:661`)
The ALLOWED actions say _"run code reviews via the code-reviewer subagent"_ but Agent tool is hard-blocked at runtime. Broker will attempt a forbidden action and fail. Fix: change to _"delegate code reviews to workers via pinet\_message"_.

**2. Only Agent gets runtime enforcement** (`guardrails.ts:143`)
`BROKER_FORBIDDEN_TOOLS` contains only `"Agent"`. edit/write/bash are forbidden in the prompt but have no runtime block. A jailbroken LLM could bypass. Consider: (a) document this as intentional (broker needs bash for inspection), or (b) enable `readOnly` mode for broker role to get runtime blocks on WRITE\_TOOLS too.

### 💡 Suggestion
- Add logging when `tool_call` blocks Agent — aids debugging when broker repeatedly hits the block.

### Checklist
- ✅ Correctness — runtime block logic correct, non-broker roles unaffected
- ✅ Architecture — clean separation, extensible forbidden-tools set, follows existing patterns
- ✅ Security — runtime block can't be bypassed by prompt, case sensitivity tested
- ✅ Tests — solid unit coverage for all new functions, consistent with existing test patterns

PR #158, opened 2026-04-02
fix: defer follower ACK until drain succeeds — at-least-once delivery (#136)

### LINKED ISSUE #136: bug: follower inbox can double-deliver messages on crash/restart
## Problem

The follower polling loop in `connectAsFollower()`:

```typescript
const entries = await client.pollInbox();   // 1. Fetch from broker DB
// ...
inbox.push(...synced.inboxMessages);        // 2. Push to in-memory inbox
// ...
await client.ackMessages(ids);              // 3. ACK (marks delivered in DB)
```

If the follower crashes between step 2 and step 3:
- Messages are in the in-memory inbox (lost on crash)
- Messages are NOT acked in the broker DB
- On restart, the follower polls again → same messages delivered again

If the crash happens after step 3 but before the agent processes `drainInbox()`:
- Messages are acked (won't be redelivered)
- Messages are in the in-memory inbox (lost on crash)
- **Messages are lost permanently**

## Impact

- Message loss on crash after ACK
- Message duplication on crash before ACK
- Neither behavior is documented or handled

## Suggested Fix

Option A: ACK-on-delivery — only ACK after the agent actually processes the message (not just polls it). This prevents loss but allows duplication.

Option B: Persist inbox to disk before ACK — guarantees at-least-once delivery even across crashes.

At minimum, document the delivery guarantee (at-most-once currently) and make it explicit.

## Severity

**High** — message loss is the worst failure mode for a coordination system.

gugu91:
## Problem

The follower polling loop had an ACK-before-drain race condition:

```
poll → push to inbox → ACK → [crash here] → drain never happens
```

If the follower crashed between ACK and `drainInbox()`, messages were **permanently lost** — the broker thought they were delivered, but the agent never saw them.

A second issue: if the poll interval fired before drain, the same messages would be double-pushed to the inbox (since they weren't ACK'd yet in the old code, but there was no dedup either).

## Fix

**At-least-once delivery semantics** via deferred ACK:

1. **`pendingFollowerAckIds`** — new array tracks inbox IDs that have been polled but not yet drained
2. **Poll loop** — pushes messages to inbox and records IDs, but does **not** ACK. A dedup filter (Set check) prevents double-push when poll fires before drain.
3. **`drainInbox()`** — after `sendUserMessage()` succeeds, calls `ackPendingFollowerMessages()` to ACK the pending IDs. If delivery fails, messages go back to inbox and IDs stay pending.
4. **Clean shutdown** — flushes pending ACKs before disconnect (agent has already processed them)

### Failure modes now:

| Scenario | Outcome |
|----------|---------|
| Crash before drain | IDs not ACK'd → broker re-delivers on reconnect ✅ |
| Crash after drain, before ACK | Messages processed, ACK lost → broker re-delivers (duplicate) → acceptable ✅ |
| Drain fails (sendUserMessage throws) | Messages back in inbox, IDs stay pending → retry on next drain ✅ |
| Poll fires before drain | Dedup filter skips already-pending IDs → no double-push ✅ |
| Clean shutdown | Pending ACKs flushed → no unnecessary re-delivery ✅ |

## Testing

- `pnpm lint` ✅
- `pnpm typecheck` ✅
- `pnpm test` ✅ (394 tests pass)

## Files changed

- `slack-bridge/index.ts` — 53 insertions, 3 deletions

Closes #136

### DIFF
--- a/slack-bridge/index.ts
+++ b/slack-bridge/index.ts
@@ -1227,6 +1227,12 @@ export default function (pi: ExtensionAPI) {
   let activeBroker: any = null;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   let brokerClient: any = null;
+
+  // Deferred ACK tracking for follower mode.
+  // IDs are collected on poll and only ACK'd after drainInbox() successfully
+  // delivers messages to the agent. This prevents message loss when the
+  // follower crashes between ACK and drain (at-least-once semantics).
+  const pendingFollowerAckIds: number[] = [];
   let activeRouter: MessageRouter | null = null;
   let activeSelfId: string | null = null;
   let brokerHeartbeatTimer: ReturnType<typeof setInterval> | null = null;
@@ -1792,7 +1798,13 @@ export default function (pi: ExtensionAPI) {
           const entries = await client.pollInbox();
           if (entries.length === 0) return;
 
-          const synced = syncFollowerInboxEntries(entries, threads, agentName, lastDmChannel);
+          // Dedup: skip entries already pending ACK (polled but not yet drained).
+          // This prevents double-push when the poll interval fires before drain.
+          const pendingSet = new Set(pendingFollowerAckIds);
+          const newEntries = entries.filter((entry) => !pendingSet.has(entry.inboxId));
+          if (newEntries.length === 0) return;
+
+          const synced = syncFollowerInboxEntries(newEntries, threads, agentName, lastDmChannel);
           for (const nextThread of synced.threadUpdates) {
             const existing = threads.get(nextThread.threadTs);
             if (!existing) {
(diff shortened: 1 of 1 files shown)


will-jack-and-jilll (changes requested):
*Will's reviewer agent Queue Otter*

## Verdict
- **Block**
- The deferred-ACK idea is correct, but the new ACK state machine still has one message-loss path and one duplicate-delivery race.

## Top findings
- `slack-bridge/index.ts:2149` — `session_shutdown` ACKs every pending follower ID, even though `pendingFollowerAckIds` includes messages that may still be sitting in `inbox[]` waiting for `drainInbox()`. If the follower polls while the agent is busy and the session exits before drain, those messages are marked delivered and then lost.
- `slack-bridge/index.ts:2055` — `ackPendingFollowerMessages()` removes IDs from the dedup source before `ackMessages()` resolves. A subsequent `pollInbox()` can therefore re-read the same unacked rows while the ACK is still in flight, re-enqueueing duplicates.

## Suggested fix
- Split the lifecycle into distinct states, e.g. `queuedButUndelivered`, `deliveredAwaitingAck`, and `ackInFlight`.
- Only ACK IDs that were successfully handed to `pi.sendUserMessage()`.
- Keep dedup active until the ACK round-trip finishes, and add regression tests for shutdown-before-drain and slow/failed ACKs.

PR #163, opened 2026-04-02
fix: harden broker JSON-RPC request validation (#147)

### LINKED ISSUE #147: bug: broker socket server onConnection doesn't validate JSON-RPC version
## Problem

In `socket-server.ts`, the `processBuffer()` method validates:
```typescript
if (request.jsonrpc \!== '2.0' || typeof request.method \!== 'string' || request.id === undefined) {
  this.send(socket, rpcError(null, RPC_INVALID_REQUEST, 'Invalid request'));
  continue;
}
```

But `request.id === undefined` allows `id: null`, `id: 0`, `id: ''`, and `id: false` — all of which are valid JSON-RPC notification IDs but shouldn't be used for request-response.

More importantly, there's **no authentication** on the socket. Any process that can connect to `~/.pi/pinet.sock` can:
- Register as an agent
- Poll any agent's inbox
- Send messages as any agent
- Claim any thread

This is a local-only socket so the risk is limited, but it means any local process (or malicious extension) can impersonate agents.

## Suggested Fix

1. Fix the ID validation to reject falsy values (JSON-RPC allows null only for notifications)
2. Consider adding a shared secret or capability token at registration time
3. At minimum, validate that the socket permissions are user-only (`0600`)

## Severity

**Low** — local-only attack surface, but worth hardening.

gugu91:
## Problem

`BrokerSocketServer.processBuffer()` parsed JSON and then only checked:

- `request.jsonrpc === "2.0"`
- `typeof request.method === "string"`
- `request.id \!== undefined`

That left a few gaps:
- non-object payloads like `null`, `[]`, `true`, `123` could make validation brittle
- invalid IDs like `null`, `false`, `0`, and `""` were accepted
- malformed requests with invalid `params` shapes were not rejected early

## Fix

Added explicit request validation helpers in `socket-server.ts`:
- `isJsonRpcRequestId()`
- `isJsonRpcRequestPayload()`
- `extractJsonRpcRequestId()`

The socket server now rejects invalid requests before dispatch unless all of these are true:
- payload is a non-null object (not an array)
- `jsonrpc === "2.0"`
- `method` is a non-empty string
- `id` is a valid request ID for this broker protocol
- `params`, if present, is an object (not an array / primitive / null)

Also improved invalid-request responses to echo a valid request ID when one can be safely extracted.

## Tests

Added broker socket-server coverage for:
- parse errors on invalid JSON
- invalid non-object payloads
- wrong JSON-RPC version
- invalid request IDs
- non-object params

## Verification

- ✅ `pnpm lint`
- ✅ `pnpm typecheck`
- ✅ `pnpm test` (398 passing)

## Scope note

This PR fixes the JSON-RPC validation bug from #147. It does **not** add socket auth/shared-secret protection yet.

Closes #147

### DIFF
--- a/slack-bridge/broker/helpers.test.ts
+++ b/slack-bridge/broker/helpers.test.ts
@@ -455,9 +455,7 @@ describe("BrokerDB", () => {
 
     const sqlite = (db as unknown as { getDb(): DatabaseSync }).getDb();
     sqlite
-      .prepare(
-        "UPDATE agents SET disconnected_at = ?, resumable_until = NULL WHERE id = ?",
-      )
+      .prepare("UPDATE agents SET disconnected_at = ?, resumable_until = NULL WHERE id = ?")
       .run(new Date(Date.now() - 2 * 60 * 60_000).toISOString(), "gone");
 
     runBrokerMaintenancePass(db, {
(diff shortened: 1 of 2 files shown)


will-jack-and-jilll (approved):
*Will's reviewer agent Panic Ferret*\n\n## Verdict\n- **Approve**\n- I reviewed the JSON-RPC validation hardening in `slack-bridge/broker/socket-server.ts` and the added coverage in `slack-bridge/broker/helpers.test.ts`. I did not find any blocking correctness, regression, or maintainability issues in this diff.\n\n## Findings\n- None.\n\n## Notes\n- The server now rejects malformed non-object payloads, wrong `jsonrpc` versions, invalid request IDs, and non-object `params` before dispatch.\n- The added tests cover the risky invalid-request paths well.\n- I did not rerun local automation here because the PR head is not checked out in this read-only review environment.

PR #168, opened 2026-04-02
fix: warn when SQLite WAL mode falls back (#142)

### LINKED ISSUE #142: bug: SQLite WAL mode silently falls back if it can't be set
## Problem

Both `schema.ts` (BrokerDB) and `comments-sqlite.ts` (SqliteCommentStore) execute:

```typescript
db.exec('PRAGMA journal_mode=WAL');
```

But SQLite doesn't error if WAL mode can't be enabled (e.g., if another process holds a lock, or the filesystem doesn't support it). It silently falls back to DELETE mode.

In DELETE mode:
- Concurrent reads block writes
- No read-write concurrency
- The broker's maintenance timer, ralph loop, and socket server all hit the same DB simultaneously
- This can cause `SQLITE_BUSY` errors even with `busy_timeout`

The code doesn't check the return value of the PRAGMA.

## Suggested Fix

```typescript
const result = db.prepare('PRAGMA journal_mode=WAL').get();
if (result?.journal_mode \!== 'wal') {
  console.warn('[BrokerDB] WAL mode not available, using fallback');
}
```

Also consider logging the current journal mode at startup.

## Severity

**Low-Medium** — only manifests under specific filesystem conditions, but hard to debug when it does.

gugu91:
## Problem

Both `BrokerDB` and `SqliteCommentStore` used:

```ts
db.exec("PRAGMA journal_mode=WAL")
```

SQLite does not throw if WAL cannot be enabled. It silently returns the actual journal mode instead, so the code failed open with no signal when it fell back to `delete` or another mode.

That makes WAL-related concurrency problems much harder to diagnose.

## Fix

Switched both call sites to inspect the PRAGMA result directly:

- `slack-bridge/broker/schema.ts`
- `nvim-bridge/comments-sqlite.ts`

If the returned journal mode is not `wal`, we now emit an explicit warning including the actual fallback mode.

I also added small helper functions + tests in `slack-bridge/helpers.ts` / `helpers.test.ts` for parsing and warning on journal-mode fallback.

## Checks

- `pnpm lint` ✅
- `pnpm typecheck` ✅
- `pnpm test` ✅ 397 passing

Closes #142

### DIFF
--- a/nvim-bridge/comments-sqlite.ts
+++ b/nvim-bridge/comments-sqlite.ts
@@ -44,6 +44,15 @@ interface IdRow {
   id: string;
 }
 
+interface SqliteJournalModeResult {
+  journal_mode?: string | null;
+}
+
+function getSqliteJournalMode(result?: SqliteJournalModeResult): string {
+  const mode = result?.journal_mode?.trim().toLowerCase();
+  return mode && mode.length > 0 ? mode : "unknown";
+}
+
 function rowToCommentRecord(row: CommentRow): CommentRecord {
   const context: CommentContext | undefined =
     row.context_file != null
@@ -84,7 +93,14 @@ export class SqliteCommentStore implements ICommentStore {
 
     this.db = new DatabaseSync(this.dbPath, { timeout: 5000 });
 
-    this.db.exec("PRAGMA journal_mode=WAL");
+    const journalMode = this.db.prepare("PRAGMA journal_mode=WAL").get() as
+      | SqliteJournalModeResult
+      | undefined;
+    if (getSqliteJournalMode(journalMode) !== "wal") {
+      console.warn(
+        `[SqliteCommentStore] SQLite WAL mode not available, using ${getSqliteJournalMode(journalMode)} journal mode fallback`,
+      );
+    }
 
     this.db.exec(`
       CREATE TABLE IF NOT EXISTS comments (
--- a/slack-bridge/broker/schema.ts
+++ b/slack-bridge/broker/schema.ts
@@ -12,6 +12,11 @@ import type {
   InboundMessage,
   ChannelAssignment,
 } from "./types.js";
+import {
+  buildSqliteWalFallbackWarning,
+  isSqliteWalEnabled,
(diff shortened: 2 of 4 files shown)


will-jack-and-jilll (approved):
*Will's reviewer agent Pragma Panda*

## Verdict
- **Approve**
- Small, targeted fix and no blockers found. The PR now checks SQLite's actual `journal_mode` result instead of assuming WAL was enabled, which closes the previous fail-open behavior and improves diagnosability when SQLite falls back to another mode.

## Review Scope
- PR reviewed: `#168 https://github.com/gugu91/extensions/pull/168`
- Files examined: `nvim-bridge/comments-sqlite.ts`, `slack-bridge/broker/schema.ts`, `slack-bridge/helpers.ts`, `slack-bridge/helpers.test.ts`
- Commands run: `gh pr view`, `gh pr diff`, `gh api repos/.../pulls/168/files`, `gh api repos/.../contents/...`, `gh pr checks`, `node --input-type=module -e ...`
- PiComms: `pr-168-review` comment posted
- GitHub: approving review posted

## Findings

### Critical (must fix)
- None.

### Warnings (should fix)
- None.

### Suggestions (consider)
- None.

## Checklist
- Correctness: pass — both WAL call sites now inspect SQLite's returned mode and warn on fallback.
- Architecture & maintainability: pass — helper extraction in `slack-bridge` keeps the new parsing/warning logic straightforward and testable.
- Security & reliability: pass — the change improves runtime visibility for a silent SQLite fallback that can affect concurrency behavior.
- Tests & verification: pass — helper tests were added, and I manually verified with `node:sqlite` that `PRAGMA journal_mode=WAL` returns the effective mode; GitHub reported no branch checks.

## Summary
- No correctness or regression issues stood out in the diff.
- The fix addresses the real SQLite behavior: WAL requests can silently fall back, and the code now surfaces that explicitly.
- `slack-bridge` added focused helper tests for mode parsing and warning text.
- Approval is based on diff review plus targeted `node:sqlite` verification; GitHub showed no automated checks for this branch.

PR #177, opened 2026-04-02
fix: stop replaying stale RALPH ghost alerts

### LINKED ISSUE #128: fix: RALPH loop keeps firing stale ghost IDs — never clears reaped agents

gugu91:
Closes #128

Supersedes #165.

## Summary
- extract ghost anomaly rewrite into a helper that reports only ghost deltas
- keep persistent ghosts out of repeated RALPH warnings until something actually changes
- suppress misleading recovery/follow-up noise while a ghost still exists
- add a multi-cycle regression test for detect → persist → unrelated anomaly → clear

## Verification
- pnpm lint
- pnpm typecheck
- pnpm test

### DIFF
--- a/slack-bridge/helpers.test.ts
+++ b/slack-bridge/helpers.test.ts
@@ -15,6 +15,7 @@ import {
   buildAgentDisplayInfo,
   rankAgentsForRouting,
   evaluateRalphLoopCycle,
+  rewriteRalphLoopGhostAnomalies,
   buildRalphLoopNudgeMessage,
   buildRalphLoopAnomalySignature,
   buildRalphLoopFollowUpMessage,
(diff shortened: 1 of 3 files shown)


will-jack-and-jilll (approved):

## Verdict
- **Approve**
- The change matches the PR intent: it suppresses repeated ghost-only RALPH noise while still surfacing new ghosts, non-ghost anomalies, and ghost clear events. I did not find a correctness or regression issue in the implementation itself.

## Review Scope
- PR reviewed: `#177 https://github.com/gugu91/extensions/pull/177`
- Files examined: `slack-bridge/index.ts`, `slack-bridge/helpers.ts`, `slack-bridge/helpers.test.ts`
- Commands run: `gh pr view 177 --json ...`, `gh pr diff 177 --name-only`, `gh pr diff 177 --patch`, `git show <commit>:...`, `gh pr checks 177`
- PiComms: `pr-177-review` comment posted
- GitHub: summary review posted

## Findings

### Critical (must fix)
- None.

### Warnings (should fix)
- None.

### Suggestions (consider)
- `slack-bridge/helpers.test.ts:1036` — Consider adding one more regression test for multiple ghosts changing membership across cycles (for example `ghost-1` persists while `ghost-2` appears, then only `ghost-1` clears).
  - Why it may help
    - The helper is now stateful across cycles, so one mixed add/remove case would make the delta semantics more explicit and guard the most complex edge case.

## Checklist
- Correctness: pass — Delta rewrite logic and notification gating line up with the stated behavior.
- Architecture & maintainability: pass — The helper extraction keeps the policy localized instead of spreading string-rewrite logic through the broker loop.
- Security & reliability: pass — No new privileged behavior; failure paths remain best-effort and bounded.
- Tests & verification: pass — The new multi-cycle regression test covers the core detect → persist → unrelated anomaly → clear flow. I did not run the suite locally because the review target was the GitHub PR commit rather than the checked-out worktree.

## Summary
- Ghost-only RALPH warnings are now deduped by state delta instead of replaying every cycle.
- Non-ghost anomalies still surface while a ghost persists, which preserves actionable signals.
- Ghost clear events are surfaced without sending a misleading full recovery/follow-up message.
- Test coverage matches the main behavioral change and exercises the intended multi-cycle regression path.

PR #186, opened 2026-04-02
fix: harden follower inbox delivery across restart

### LINKED ISSUE #136: bug: follower inbox can double-deliver messages on crash/restart

gugu91:
## Summary
- replace the PR #158 approach with a fresh follower delivery state machine that separates queued, delivered, and ack-in-flight inbox IDs
- defer follower ACKs until messages are actually handed to `pi.sendUserMessage()`, keep dedup active until the ACK round-trip completes, and only flush delivered IDs on shutdown
- add focused follower delivery tests plus broker inbox ID coverage in `syncFollowerInboxEntries`

## Why fresh instead of salvaging PR #158
PR #158 had the right direction but the review correctly pointed out two remaining races:
- queued-but-undrained messages could still be ACKed on shutdown and lost
- IDs stopped being deduped before `ackMessages()` finished, so a slow ACK could re-enqueue duplicates

This PR keeps the same at-least-once goal but reworks the lifecycle to address those review findings directly.

## Validation
- `pnpm lint`
- `pnpm typecheck`
- `pnpm test`

Closes #136

### DIFF
--- a/slack-bridge/follower-delivery.test.ts
+++ b/slack-bridge/follower-delivery.test.ts
@@ -1,0 +1,100 @@
+import { describe, expect, it } from "vitest";
+import {
+  createFollowerDeliveryState,
+  drainFollowerAckBatches,
+  getFollowerShutdownAckIds,
+  hasDeliveredFollowerInboxIds,
+  isFollowerInboxIdTracked,
+  markFollowerAckBatchFailed,
+  markFollowerAckBatchSucceeded,
+  markFollowerInboxIdsDelivered,
+  queueFollowerInboxIds,
+  resetFollowerDeliveryState,
+  takeFollowerAckBatch,
+} from "./follower-delivery.js";
+
+describe("follower delivery state", () => {
+  it("keeps queued messages out of shutdown ACKs", () => {
+    const state = createFollowerDeliveryState();
+
+    queueFollowerInboxIds(state, [101, 102]);
+    markFollowerInboxIdsDelivered(state, [201]);
+
+    expect(getFollowerShutdownAckIds(state)).toEqual([201]);
+    expect(isFollowerInboxIdTracked(state, 101)).toBe(true);
+    expect(isFollowerInboxIdTracked(state, 102)).toBe(true);
+    expect(isFollowerInboxIdTracked(state, 201)).toBe(true);
+  });
+
+  it("keeps IDs tracked while ACK is in flight and clears them only on success", () => {
+    const state = createFollowerDeliveryState();
+
+    queueFollowerInboxIds(state, [301]);
+    markFollowerInboxIdsDelivered(state, [301]);
+
+    const batch = takeFollowerAckBatch(state);
+    expect(batch).toEqual([301]);
+    expect(isFollowerInboxIdTracked(state, 301)).toBe(true);
+    expect(hasDeliveredFollowerInboxIds(state)).toBe(false);
+
+    markFollowerAckBatchSucceeded(state, batch);
+    expect(isFollowerInboxIdTracked(state, 301)).toBe(false);
+  });
+
+  it("restores delivered IDs after an ACK failure so they can be retried", () => {
+    const state = createFollowerDeliveryState();
+
+    queueFollowerInboxIds(state, [401, 402]);
+    markFollowerInboxIdsDelivered(state, [401, 402]);
+
+    const batch = takeFollowerAckBatch(state);
+    markFollowerAckBatchFailed(state, batch);
+
+    expect(hasDeliveredFollowerInboxIds(state)).toBe(true);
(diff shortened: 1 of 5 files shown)


will-jack-and-jilll (changes requested):

## Verdict
- **Block**
- The delivery-state refactor is moving in the right direction, but there is still a shutdown race that can leave a second ACK batch in flight while teardown unregisters/disconnects the follower. That means already-delivered messages can still be replayed after restart, which undercuts the PR’s stated goal of fixing double-delivery across restart.

## Review Scope
- PR reviewed: `#186 https://github.com/gugu91/extensions/pull/186`
- Files examined: `slack-bridge/index.ts`, `slack-bridge/follower-delivery.ts`, `slack-bridge/follower-delivery.test.ts`, `slack-bridge/helpers.ts`, `slack-bridge/helpers.test.ts`
- Commands run: `gh pr view 186 --json ...`, `gh pr diff 186 --patch`, `gh pr checks 186`, `gh api repos/gugu91/extensions/pulls/186/files`, `git show <commit>:...`
- PiComms: `pr-186-review` comment posted
- GitHub: `review posted (request changes)`

## Findings

### Critical (must fix)
- `slack-bridge/index.ts:1781` / `slack-bridge/index.ts:1906` — shutdown does not reliably wait for chained follower ACK batches
  - Why it matters
  - `flushDeliveredFollowerAcks()` only tracks the current raw `ackMessages()` promise. If batch A is in flight and another batch becomes deliverable before A finishes, the function kicks off batch B with `void flushDeliveredFollowerAcks()` after A succeeds. During `session_shutdown`, the code awaits `followerAckPromise` once, then only calls `flushDeliveredFollowerAcks()` when `deliveredAwaitingAckIds` is non-empty. If batch B has already been moved into `ackInFlightIds`, shutdown sees no delivered IDs and proceeds to `unregister()` / `disconnect()` while batch B is still running.
  - Those messages were already handed to `pi.sendUserMessage()`, so aborting their ACK means they can be replayed on the next startup/reconnect — exactly the duplicate-after-restart case this PR is trying to eliminate.
  - Suggested fix
  - Make ACK draining a single awaitable operation that covers the whole chain, not just the current transport call. For example, loop inside `flushDeliveredFollowerAcks()` until there are no `deliveredAwaitingAckIds` and no `ackInFlightIds`, or track a higher-level flush promise and await that from shutdown. Add a regression test for “batch A in flight, batch B delivered before shutdown” so teardown cannot exit early.

### Warnings (should fix)
- `slack-bridge/follower-delivery.test.ts` — missing regression coverage for the shutdown/ack-chaining race
  - Why it matters
  - The new unit tests validate the set transitions, but they do not cover the exact lifecycle bug this PR is addressing: messages becoming deliverable while an earlier ACK is still in flight, then shutdown/reconnect happening in between.
  - Suggested fix
  - Add an integration-style test around the follower polling/drain/teardown flow, or at least a targeted test for the ack-draining helper that proves shutdown waits for every chained batch.

### Suggestions (consider)
- None.

## Checklist
- Correctness: fail with a remaining shutdown race in ACK chaining
- Architecture & maintainability: pass with a cleaner delivery-state abstraction
- Security & reliability: fail because teardown can still reintroduce duplicate delivery
- Tests & verification: fail because the chained-ACK shutdown path is not exercised

## Summary
- The refactor improves state tracking and fixes part of the earlier ACK timing problem.
- A remaining shutdown race means a second ACK batch can still be interrupted after delivery.
- That can reintroduce duplicate delivery after restart, so I’m blocking until teardown waits for the full ACK chain.
- Please add a regression test for this exact sequence once fixed.

PR #242, opened 2026-04-03
fix: skip broker in RALPH idle nudges

### LINKED ISSUE #241: bug: RALPH loop nudges broker as idle-with-assigned-work
**Priority: P0**

RALPH flags the broker agent itself as idle with assigned work and nudges it to pick up backlog:

```
RALPH loop (2026-04-03T09:11:33.959Z): Hyper Burgundy Hedgehog idle with assigned work (1 inbox, 3 threads); pending backlog (3) with 29 idle workers
```

The broker is **not** a worker — it coordinates, routes messages, and runs RALPH. It should never be nudged to pick up implementation work. This is a false positive that adds noise to every RALPH cycle.

**Expected**: broker agent is excluded from the idle-with-assigned-work evaluation entirely (or treated as a special role that does not receive nudge messages).

**Actual**: broker nudges itself, creating a self-referential Easter egg nobody asked for 🐣

This fires every single RALPH cycle and clutters the control plane.

gugu91:
## Summary
- exclude the broker's own agent id from the RALPH idle-with-assigned-work evaluation
- thread that broker id through both RALPH evaluation call sites so the live loop and dashboard agree
- add a regression test proving the broker is not nudged even if it would otherwise look idle with assigned work

## Testing
- `cd .worktrees/fix-241 && pnpm lint && pnpm typecheck && pnpm test`

Closes #241.

### DIFF
--- a/slack-bridge/helpers.test.ts
+++ b/slack-bridge/helpers.test.ts
@@ -1407,6 +1407,48 @@ describe("evaluateRalphLoopCycle", () => {
     expect(result.anomalies.some((item) => item.includes("expected `main`"))).toBe(true);
   });
 
+  it("skips the broker agent id when flagging idle agents with assigned work", () => {
+    const result = evaluateRalphLoopCycle(
+      [
+        {
+          emoji: "🦙",
+          name: "Broker Llama",
+          id: "broker-self",
+          status: "idle",
+          metadata: { role: "worker" },
+          lastSeen: "2026-04-01T00:00:00.000Z",
+          lastHeartbeat: "2026-04-01T00:01:55.000Z",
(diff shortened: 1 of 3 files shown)


gugu91 (commented):
Review verdict: approve.

I verified that `brokerAgentId` is threaded through both `evaluateRalphLoopCycle(...)` call sites, the helper skips the broker self row so RALPH no longer self-nudges when that row looks like a worker, and the new regression test proves a normal worker in the same scenario is still evaluated and nudged.

I also ran `pnpm lint && pnpm typecheck && pnpm test` on `.worktrees/fix-241`; all passed. No findings.

PR #278, opened 2026-04-08
fix: expire stale targeted broker backlog

### LINKED ISSUE #271: Broker leaves stale ghost agents and undrainable backlog after disconnects
## Summary

The broker is keeping stale disconnected agents in the registry and repeatedly surfacing a persistent RALPH anomaly:

- `pending backlog (3) with 2 idle workers`

Even across multiple cycles, the backlog does not drain.

## What I observed

At `2026-04-08T08:49:11Z` and subsequent cycles:
- many ghost agent IDs were still reported
- `ralph_cycles.backlog_count = 3`
- anomaly remained `pending backlog (3) with 2 idle workers`

Read-only inspection of `~/.pi/pinet-broker.db` shows:
- stale disconnected agents from prior `extensions` sessions still present in `agents`
- three old `unrouted_backlog` rows remain from `2026-04-03`
- those rows are agent-to-agent messages tied to dead/disconnected broker-era agents
- at least some pending backlog entries point at a missing preferred agent (`93ab1c8e`) and never get drained

Relevant rows looked like:
- `unrouted_backlog.id in (2,3,4)` — `status='pending'`
- stale thread: `a2a:5ece9ad2-4b37-46da-826a-22048413fa1b:93ab1c8e`
- preferred target missing: `preferred_agent_id='93ab1c8e'`

## Why this matters

RALPH keeps warning about backlog + idle capacity, but the queue contains stale/disconnected-session residue rather than actionable live work. That creates false-positive maintenance pressure and makes broker health harder to trust.

## Desired behavior

- stale ghost agents should be purged on schedule
- backlog tied to dead/unresolvable agent-to-agent conversations should either:
  - be reassigned when safe, or
  - be dropped/expired with a clear reason when not safe
- RALPH should stop reporting drainable backlog when only stale residue remains

## Likely code areas

- `slack-bridge/broker/maintenance.ts`
- broker DB lifecycle / purge / requeue logic
- backlog assignment policy around `preferred_agent_id`
- any logic that preserves disconnected agents or orphaned a2a backlog too long

## Acceptance criteria

- reproducer coverage for stale disconnected agents + stale a2a backlog
- broker purges or resolves these rows deterministically
- RALPH no longer reports idle-drain anomalies for undrainable stale backlog
- tests cover the cleanup behavior

gugu91:
## Summary
- drop irrecoverable targeted A2A backlog once the intended recipient is gone
- preserve hold-and-drain semantics for targeted backlog while the intended recipient is still recoverable
- repair orphaned targeted assigned backlog rows with no inbox so they either return to pending or expire safely
- lock the acked targeted backlog path so maintenance does not requeue or drop successfully drained work

## Testing
- pnpm lint
- pnpm typecheck
- pnpm test

Closes #271

### DIFF
--- a/slack-bridge/broker/helpers.test.ts
+++ b/slack-bridge/broker/helpers.test.ts
@@ -1483,6 +1483,28 @@ describe("BrokerDB", () => {
     expect(db.getInbox("a2")).toHaveLength(1);
   });
 
+  it("markDelivered clears assigned targeted backlog after the intended recipient acks it", () => {
+    db.registerAgent("sender", "Sender", "📤", 1);
+    db.registerAgent("target", "Target", "📥", 2);
+    db.createThread("a2a:sender:target", "agent", "", "sender");
+    db.insertMessage("a2a:sender:target", "agent", "inbound", "sender", "complete me", ["target"], {
+      senderAgent: "Sender",
+      a2a: true,
+    });
+
+    expect(db.requeueUndeliveredMessages("target")).toBe(1);
+    const [backlog] = db.getPendingBacklog();
+    expect(db.assignBacklogEntry(backlog.id, "target")?.status).toBe("assigned");
+
+    const inbox = db.getInbox("target");
+    expect(inbox).toHaveLength(1);
+
+    db.markDelivered([inbox[0].entry.id], "target");
+
+    expect(db.getInbox("target")).toHaveLength(0);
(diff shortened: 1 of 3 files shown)


gugu91:
Superseded by #284, which isolates the one essential follow-up missing from merged main: finalize targeted assigned backlog on ack and lock the deliver/ack -> purge -> maintenance regression.

PR #347, opened 2026-04-09
fix: debounce fresh RALPH ghost alerts (#331)

gugu91:
## Summary
- suppress freshly reaped heartbeat-timeout ghosts from the same RALPH cycle that reaped them
- only surface them as `NEW ghost agents detected` if they are still ghosted in a later cycle
- add regression coverage for transient reaps vs persistent ghosts

## Root cause
RALPH currently does this in one pass:
1. broker maintenance runs first and `pruneStaleAgents()` can mark overdue heartbeat rows disconnected
2. the same RALPH cycle immediately evaluates those rows as ghosts
3. ghost rewrite treats every never-before-reported ghost id as `NEW`

That means a transient broker/heartbeat timing race can produce this bad shape:
- a healthy connected worker crosses the stale-heartbeat cutoff just long enough to get reaped
- RALPH announces it as a **NEW ghost** in that same cycle
- the worker heartbeats/reconnects immediately after, so follow-up inspection shows it healthy again
- if the same transient reap happens later, it can be re-announced as `NEW` again because the ghost state cleared between cycles

## Fix
- pass the current maintenance pass `reapedAgentIds` into ghost-anomaly rewriting as suppressed ghost ids for that exact cycle
- keep those freshly reaped ids out of the visible ghost set / `NEW ghost` alert path until they survive into a later cycle
- preserve already-reported ghost memory so suppression does not manufacture false clear/re-announce churn

## Why this is narrow and safe
- no schema change
- no routing change
- no broker pruning change
- real ghost rows still get reported if they remain ghosted after the transient reap window
- immediate maintenance logging still records the reap, so operators still see that something happened

## Testing
- `pnpm --filter @gugu910/pi-slack-bridge test -- helpers.test.ts`
- `pnpm lint`
- `pnpm typecheck`
- `pnpm test`

### DIFF
--- a/slack-bridge/helpers.test.ts
+++ b/slack-bridge/helpers.test.ts
@@ -2099,6 +2099,40 @@ describe("rewriteRalphLoopGhostAnomalies", () => {
     expect(cycle4.evaluation.anomalies).toEqual(["ghost agents cleared from registry: ghost-1"]);
     expect(cycle4.clearedGhostIds).toEqual(["ghost-1"]);
   });
+
+  it("suppresses freshly reaped ghost ids until they survive a later cycle", () => {
+    const cycle1 = rewriteRalphLoopGhostAnomalies(
+      buildEvaluation(["ghost-1", "ghost-2"], ["ghost agents detected: ghost-1, ghost-2"]),
+      [],
+      { suppressedGhostIds: ["ghost-1"] },
+    );
+
+    expect(cycle1.evaluation.ghostAgentIds).toEqual(["ghost-2"]);
+    expect(cycle1.evaluation.anomalies).toEqual(["NEW ghost agents detected: ghost-2"]);
+    expect(cycle1.nextReportedGhostIds).toEqual(["ghost-2"]);
+
+    const cycle2 = rewriteRalphLoopGhostAnomalies(
+      buildEvaluation(["ghost-1", "ghost-2"], ["ghost agents detected: ghost-1, ghost-2"]),
+      cycle1.nextReportedGhostIds,
+    );
+
+    expect(cycle2.evaluation.ghostAgentIds).toEqual(["ghost-1", "ghost-2"]);
+    expect(cycle2.evaluation.anomalies).toEqual(["NEW ghost agents detected: ghost-1"]);
+    expect(cycle2.nextReportedGhostIds).toEqual(["ghost-1", "ghost-2"]);
+  });
+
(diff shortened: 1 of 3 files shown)


will-jack-and-jilll (commented):
*Will's reviewer agent Cache Otter*\n\n## Verdict\n- **Approve**\n- I used the normal review flow against GitHub PR #347 (`gh pr status`, `gh pr view`, `gh pr diff`), then read the changed files and ran targeted read-only verification. The claimed root cause matches the observed false-positive mass-ghost behavior: broker maintenance reaps stale rows first, and the same RALPH cycle was immediately reclassifying those freshly reaped rows as `NEW ghost` alerts.\n\n## Review Scope\n- PR reviewed: `#347 https://github.com/gugu91/extensions/pull/347`\n- Files examined: `slack-bridge/helpers.ts`, `slack-bridge/ralph-loop.ts`, `slack-bridge/helpers.test.ts`, plus maintenance/schema context\n- Commands run: `gh pr status`; `gh pr view 347 --json ...`; `gh pr diff 347 --patch`; `git show`; `git grep`; `pnpm --filter @gugu910/pi-slack-bridge test -- helpers.test.ts`; `pnpm --filter @gugu910/pi-slack-bridge typecheck`\n- PiComms: `pr-347-review` comment posted\n- GitHub: review comment posted\n\n## Findings\n\n### Critical (must fix)\n- None.\n\n### Warnings (should fix)\n- None.\n\n### Suggestions (consider)\n- `slack-bridge/ralph-loop.ts:183` — A future integration-style test around `runRalphLoopCycle` would be nice to lock the maintenance-to-rewrite plumbing in place, though the current helper-level regression coverage is already solid.\n  - Why it may help\n  - It would catch accidental breakage in the call-site wiring, not just in the pure rewrite helper.\n\n## Checklist\n- Correctness: pass — same-cycle reap suppression addresses the specific race without hiding later persistent ghosts.\n- Architecture & maintainability: pass — narrow helper option, no schema or routing churn.\n- Security & reliability: pass — reduces noisy false positives without widening operational risk.\n- Tests & verification: pass — targeted regressions and typecheck passed.\n\n## Summary\n- Root cause matches the code path and the reported symptom.\n- Suppressing freshly reaped ids from the same RALPH cycle is the right narrow fix.\n- Previously reported ghosts are retained through temporary suppression, so there is no false clear/re-announce churn.\n- Merge-ready from my review perspective.

PR #366, opened 2026-04-11
refactor: preserve transport source through thread ownership

gugu91:
## Summary
- plumb transport `source` through broker thread claims instead of hardcoding Slack
- preserve that `source` in follower thread cache + owned-thread reclaim on resume
- keep Slack explicit at the adapter edge while leaving the seam small and transport-oriented

## Why
This is a narrow ports/adapters seam for the future Symphony split: it removes a core broker assumption that every claimed or resumed thread is Slack, without folding in package moves or broader Slack UX changes.

## Checks
- `pnpm lint`
- `pnpm typecheck`
- `pnpm test`
- `cd slack-bridge && pnpm test -- helpers.test.ts broker/router.test.ts broker/integration.test.ts`

## Stack / release handling
- Stacked on #333 via base branch `fix/323-slack-allowlist-defaults`
- Scope frozen to the transport-source ownership seam
- Queue behind #333; only pull into tonight's release if there's still room after the gate clears

## Release note
Preserve message transport source through broker thread ownership so future non-Slack adapters can claim and resume threads without being recorded as Slack.

### DIFF
--- a/neon-psql/python/sitecustomize.py
+++ b/neon-psql/python/sitecustomize.py
@@ -2,19 +2,38 @@ import os
 from urllib.parse import parse_qsl, urlencode, urlsplit, urlunsplit
 
 
+LOCAL_TUNNEL_HOSTS = {"127.0.0.1", "localhost"}
+
+
+def _is_local_tunnel_host(host):
+    return isinstance(host, str) and host.strip().lower() in LOCAL_TUNNEL_HOSTS
+
+
+def _dsn_uses_local_tunnel(dsn, tunnel_port):
+    if not isinstance(dsn, str) or not tunnel_port:
+        return False
+
+    try:
+        parts = urlsplit(dsn)
+    except ValueError:
+        return False
+
+    return _is_local_tunnel_host(parts.hostname) and parts.port is not None and str(parts.port) == tunnel_port
+
+
 def _uses_local_tunnel(args, kwargs):
     tunnel_port = os.getenv("NEON_TUNNEL_PORT", "")
+    if not tunnel_port:
+        return False
+
     host = kwargs.get("host")
     port = kwargs.get("port")
     dsn = kwargs.get("dsn") if isinstance(kwargs.get("dsn"), str) else (args[0] if args and isinstance(args[0], str) else None)
 
-    if isinstance(host, str) and host in {"127.0.0.1", "localhost"}:
+    if _is_local_tunnel_host(host) and port is not None and str(port) == tunnel_port:
         return True
-    if tunnel_port and str(port) == tunnel_port:
-        return True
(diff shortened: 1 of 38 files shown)


will-jack-and-jilll:
*Will's reviewer agent Socket Ferret*

## Verdict
- **Block**
- The source-plumbing change is directionally right, but there is still one upgrade/resume regression: follower-owned thread caches restored from pre-change state can still be reclaimed as `slack`, which breaks the narrow transport seam this PR is trying to preserve.

## Review Scope
- PR reviewed: `#366 https://github.com/gugu91/extensions/pull/366`
- Files examined: `slack-bridge/broker/client.ts`, `slack-bridge/broker/router.ts`, `slack-bridge/broker/socket-server.ts`, `slack-bridge/helpers.ts`, `slack-bridge/index.ts`, related tests
- Commands run: `gh pr view 366 --json ...`, `gh pr diff 366 --patch`, `gh pr view 366 --comments`, targeted `git show`/`git grep` on PR head SHA
- PiComms: `pr-366-review` comment posted
- GitHub: summary comment posted

## Findings

### Critical (must fix)
- `slack-bridge/helpers.ts:1924`, `slack-bridge/index.ts:3154`, `slack-bridge/index.ts:3507` — restored follower thread state can still lose transport source on reclaim
  - Why it matters: `resumeThreadClaims()` now reclaims with `thread.source ?? "slack"`, but persisted thread maps from older sessions are restored verbatim and `trackBrokerInboundThread()` only fills `source` on brand-new cache entries. If a previously-owned non-Slack thread is restored without `source`, the next reconnect reclaims it as Slack before polling has a chance to refresh metadata. That silently re-widens the seam and can misclassify non-Slack threads in the broker DB.
  - Suggested fix: when broker inbound traffic arrives, update existing cached threads if `source` is newly available; and/or add an explicit migration/fallback path so restored threads without `source` are not reclaimed as Slack until the broker has confirmed their source. A regression test should cover restoring old persisted state and reconnecting before inbox sync.

### Warnings (should fix)
- None.

### Suggestions (consider)
- None.

## Checklist
- Correctness: **fail** — reconnect/restore path can still stamp non-Slack ownership back to Slack.
- Architecture & maintainability: **fail** — the seam stays mostly narrow, but the persisted-cache fallback leaks a Slack default back into transport ownership.
- Security & reliability: **pass** — no obvious security issue in the reviewed seam; main concern is state corruption/regression.
- Tests & verification: **fail** — existing tests cover fresh source propagation, but not the old-state restore/reconnect boundary that introduces the regression.

## Summary
- Good narrow seam overall: `source` is now carried through broker claims and follower-owned reclaim.
- The remaining hole is the persisted follower thread cache upgrade path.
- On reconnect, old cached threads with no `source` still default to Slack before broker sync.
- I’d merge after adding a restore/reconnect regression fix and test.

Issue #496, opened 2026-04-18
security: stop exposing ownerAgentId in broker registration conflict responses


gugu91:
## Summary
Follow-up narrow slice from #324 after #491, #493, and the in-flight #495 publication-blocked artifact: check whether raw `ownerAgentId` should still be exposed in broker registration conflict responses.

## Why
The live takeover path is now fixed and `agents.list` no longer exposes raw `stableId` values. The remaining registration conflict payload still appears to include `ownerAgentId`, which may be a separate client-visible identity surface under the broader #324 umbrella.

## Scope
- confirm the current `ownerAgentId` registration-conflict payload seam on `main`
- identify the smallest safe change to stop exposing raw `ownerAgentId` on that client-visible path, if appropriate
- preserve any still-legitimate reconnect/runtime behavior that depends on the conflict response semantics
- add focused coverage only for this payload surface if the seam is confirmed implementation-ready

## Out of scope
- broader identity-binding redesign
- broader mesh trust redesign
- the already-fixed live takeover / eviction path from #491
- the already-fixed `agents.list` exposure path from #493
- the in-flight `ownerStableId` conflict-payload slice from #495

## Related
- #324 umbrella stableId / identity surface issue
- #491 live takeover rejection slice
- #493 `agents.list` stableId exposure slice
- #495 `ownerStableId` conflict-payload slice

## Deliverable
A read-only seam check and, if safe, a narrow follow-up that removes or minimizes raw `ownerAgentId` exposure from broker registration conflict responses.

gugu91:
Read-only preflight conclusion: this slice is **implementation-ready now** if it stays pinned to the registration-conflict payload only.

Clean mergeable cut:
- remove **only** `ownerAgentId` from the client-visible registration-conflict payload at the socket boundary
- leave `ownerStableId` to **#495**
- keep message / retryable / requestedName / conflict-code behavior unchanged
- extend the existing duplicate-name integration case to assert the thrown conflict error no longer exposes `ownerAgentId`

Important scope guardrail:
- do **not** widen this slice into `ownerStableId` handling, broader conflict-payload redesign, or broader identity-binding redesign unless opened separately under the #324 umbrella

gugu91:
Merged via PR #497. Closing as satisfied by the landed ownerAgentId conflict-payload exposure fix.

Issue #493, opened 2026-04-18
security: stop exposing raw stableId values to authenticated mesh clients


gugu91:
## Summary
Follow-up narrow slice from #324 after the live takeover/peer-eviction fix in #491: stop exposing raw `stableId` values back to authenticated mesh clients unless they are still strictly required.

## Why
Current `main` still exposes raw `stableId` values through broker client surfaces such as `agents.list`. Even after live takeover rejection lands, that leaves a remaining identity-surface leak under the broader #324 umbrella.

## Observed seam
Earlier preflight on #324 confirmed:
- `slack-bridge/broker/schema.ts` includes raw `stableId` in agent data returned to clients
- `slack-bridge/broker/socket-server.ts` forwards that agent data through `agents.list`

## Scope
- identify the smallest safe way to stop exposing raw `stableId` values to authenticated mesh clients on normal broker surfaces
- preserve any still-legitimate reconnect/runtime behavior that does not require client-visible raw stableIds
- add focused coverage only for the exposure change

## Out of scope
- broader identity-binding redesign
- broader mesh trust redesign
- the already-fixed live takeover / eviction path from #491

## Related
- #324 umbrella stableId / identity surface issue
- #491 live takeover rejection slice

## Deliverable
A narrow follow-up that removes or minimizes raw client-visible `stableId` exposure on broker client surfaces, if the preflight confirms it is safe and mergeable as a standalone slice.

gugu91:
Read-only preflight conclusion: this slice is **implementation-ready now** if it stays pinned to **`agents.list` only**.

Clean mergeable cut:
- add a narrow client-safe projection at the socket boundary for `agents.list`
- stop returning raw `stableId` values on that client-visible surface
- update the client-visible broker types / `listAgents()` expectations accordingly
- leave internal DB/reconnect/router uses of `stableId` untouched

Important scope guardrail:
- current `main` still has another client-visible raw `stableId` path in the registration conflict payload (`ownerStableId`)
- that broader follow-up should stay **out of this slice** and remain under the umbrella **#324** after `agents.list` is fixed

gugu91:
Merged via PR #494. Closing as satisfied by the landed agents.list stableId exposure fix.

Issue #491, opened 2026-04-18
security: reject live stableId takeover during broker registration


gugu91:
## Summary
Follow-up narrow slice from #324: reject **live stableId takeover / peer eviction** during broker registration instead of trying to solve the entire stableId problem at once.

## Why
Current `main` still allows an authenticated mesh client to present an existing `stableId`, reuse the victim agent row, and evict the currently connected holder. That is the most acute live seam and the smallest safe next fix.

## Observed seam
Read-only preflight on #324 found:
- `slack-bridge/broker/socket-server.ts` trusts client-supplied `stableId` during `register(...)`
- `slack-bridge/broker/schema.ts` resolves that `stableId` to an existing agent row and reuses the agent id
- registration then disconnects any other socket already using that agent id

## Scope
- if a supplied `stableId` already belongs to a **currently connected different agent/socket**, reject the registration
- preserve the legitimate disconnected/reconnect path for resumable clients
- add focused runtime/integration coverage for rejection vs allowed reconnect behavior

## Out of scope
- hiding `stableId` values from clients
- full identity-binding redesign
- broader mesh trust redesign beyond this live takeover seam

## Related
- #324 umbrella stableId takeover / exposure issue

## Deliverable
A narrow behavior fix that blocks live stableId takeover / peer eviction at registration time without breaking legitimate reconnects.

gugu91:
Merged via PR #492. Closing as satisfied by the landed live stableId takeover rejection fix.

Issue #324, opened 2026-04-08
security: authenticated mesh clients can impersonate and evict peers via stableId takeover


will-jack-and-jilll:
## Summary

Once a client is authenticated on the Pinet mesh, it can take over another agent's identity by registering with that agent's `stableId`.

Because `stableId` is client-supplied on registration and exposed back out through `agents.list`, an authenticated client can:
1. enumerate live agents and their `stableId`s
2. reconnect using a victim's `stableId`
3. cause the broker to reuse the victim's persisted agent row / ID
4. disconnect the victim's real socket
5. continue operating as that agent identity

This is a serious trust-boundary problem today and a hard blocker for any multi-host mesh.

### `stableId` is exposed to authenticated clients
- `slack-bridge/broker/schema.ts`
  - `rowToAgent(...)` includes `stableId: row.stable_id`
- `slack-bridge/broker/socket-server.ts`
  - `agents.list` returns the full agent objects to any registered client

### registration accepts client-supplied `stableId`
- `slack-bridge/broker/socket-server.ts`
  - `handleRegister(...)` reads `stableId` directly from request params and passes it through to the DB
- `slack-bridge/broker/schema.ts`
  - `registerAgent(...)` looks up an existing row by `stableId`
  - if a row exists, it reuses that agent identity instead of treating the client as a distinct principal

### duplicate connection handling evicts the real holder
- `slack-bridge/broker/socket-server.ts`
  - after registration, `disconnectDuplicateConnections(agent.id, socket)` destroys any other connection already using that agent ID

## Impact

Any authenticated mesh client can effectively:
- impersonate another agent
- evict that agent from the mesh
- inherit routing identity tied to that agent row
- interfere with thread ownership and coordination
- use peer-level operations such as `thread.claim`, `status.update`, `agent.message`, and control-message delivery from the victim identity

Even if mesh auth is enabled, the mesh currently behaves like a shared-secret superuser realm rather than a set of distinct authenticated principals.

## Expected behavior

A client should not be able to assume another agent's durable identity just by presenting a guessed or enumerated `stableId`.

`stableId` should not be treated as both:
- a resumable identity handle, and
- an unauthenticated client input.

## Acceptance criteria

- do not expose resumable identity handles (`stableId` or equivalent) to ordinary mesh clients unless strictly necessary
- bind agent resumption to a broker-issued or cryptographically verifiable identity, not raw client-supplied `stableId`
[...]

PR #510, opened 2026-04-19
fix(broker): stop exposing ownerStableId in registration conflict (#495)

### LINKED ISSUE #495: security: stop exposing ownerStableId in broker registration conflict responses
## Summary
Follow-up narrow slice from #324 after #491 and #493: stop exposing raw `ownerStableId` values in broker registration conflict responses if they are not still strictly required on the client-visible surface.

## Why
The acute live takeover path is now blocked and `agents.list` no longer exposes raw `stableId` values, but current `main` still appears to include a client-visible raw `stableId` in the broker registration conflict payload (`ownerStableId`). That leaves a remaining narrow exposure seam under the broader #324 umbrella.

## Scope
- confirm the current registration-conflict payload seam
- identify the smallest safe change to stop exposing raw `ownerStableId` on that client-visible path
- preserve any still-legitimate reconnect/runtime behavior that depends on the conflict response semantics
- add focused coverage only for this payload surface if the seam is confirmed implementation-ready

## Out of scope
- broader identity-binding redesign
- broader mesh trust redesign
- the already-fixed live takeover / eviction path from #491
- the already-fixed `agents.list` exposure path from #493

## Related
- #324 umbrella stableId / identity surface issue
- #491 live takeover rejection slice
- #493 `agents.list` stableId exposure slice

## Deliverable
A narrow follow-up that removes or minimizes raw `ownerStableId` exposure from broker registration conflict responses, if the preflight confirms it is safe and mergeable as a standalone slice.

tmustier:
Narrow follow-up slice under #324 after #491, #493, and #496.

## What
Stops serializing raw `ownerStableId` into the `RPC_AGENT_NAME_CONFLICT` (`data`) payload returned by the broker when a client registers with an already-reserved agent name.

## Why
Current `main` still emits the conflicting owner's raw `stableId` on this client-visible path. Nothing in the client actually consumes it — `BrokerClient`'s conflict handler only inspects `code` / message — so it's pure exposure with no functional value. This closes the last narrow client-visible raw-stableId seam called out by the #324 umbrella, per the preflight in #495.

## Scope guardrail
Intentionally narrow:
- keeps message, `code`, `requestedName`, and `retryable` unchanged
- does **not** touch `ownerAgentId` redesign or broader identity/conflict payload work (those live separately under #324)

## Test
Adds one focused assertion to the existing `rejects duplicate explicitly requested agent names with a clear retry path` integration test, mirroring the existing `ownerAgentId` check:

```ts
expect(rpcConflictError.data as Record<string, unknown>).not.toHaveProperty("ownerStableId");
```

## Gates
`pnpm prepush` (lint + typecheck + test across the workspace) passes locally — full turbo green.

Closes #495.

### DIFF
--- a/slack-bridge/broker/integration.test.ts
+++ b/slack-bridge/broker/integration.test.ts
@@ -192,6 +192,7 @@ describe("broker integration — client ↔ server ↔ DB", () => {
       }),
     );
     expect(rpcConflictError.data as Record<string, unknown>).not.toHaveProperty("ownerAgentId");
+    expect(rpcConflictError.data as Record<string, unknown>).not.toHaveProperty("ownerStableId");
 
     client2.disconnect();
   });
--- a/slack-bridge/broker/socket-server.ts
+++ b/slack-bridge/broker/socket-server.ts
@@ -592,6 +592,8 @@ export class BrokerSocketServer {
     if (explicitNameRequest) {
       const conflict = this.db.findAgentNameConflict(finalName, candidateId, stableId);
       if (conflict) {
+        // Do not expose raw ownerStableId on this client-visible payload (#495).
+        // The message + code + requestedName are enough to drive the retry path.
         return rpcError(
           req.id,
           RPC_AGENT_NAME_CONFLICT,
@@ -599,7 +601,6 @@ export class BrokerSocketServer {
           {
             code: "AGENT_NAME_CONFLICT",
             requestedName: finalName,
-            ownerStableId: conflict.stableId,
             retryable: true,
           },
         );


tmustier:
## Automated pre-merge review (`pr-reviewer` subagent)

### Findings

Code looks good. This removes `ownerStableId` at the only `RPC_AGENT_NAME_CONFLICT` serialization site, existing in-repo consumers still key off the error code/message rather than `data.ownerStableId`, and the added integration assertion is at the right seam because the client surfaces `error.data` verbatim.

### Verdict

**correct** — No P0–P2 issues found; the patch cleanly stops the remaining `ownerStableId` leak, keeps the rest of the conflict contract intact, and stays within the issue's stated scope.

gugu91:
Thanks for the narrow follow-up. I checked this against current `main`: `slack-bridge/broker/socket-server.ts` still exposes `ownerStableId` on the client-visible registration-conflict payload today, and this PR removes exactly that field without widening into broader broker protocol or diagnostics changes. The cut stays pinned to the live `#495` seam, and the focused integration assertion covers the intended behavior.

Verdict: merge.

— Silent Slate Mantis

PR #511, opened 2026-04-19
feat(slack-bridge): add broker reload controls and visibility

tmustier:
## Summary
- add broker-side Slack reload controls with in-thread visibility and model/thinking overrides
- route remote-control settlement through a generic `onCommandSettled` callback
- fix the queued `/exit` case so reload completion and override application still happen before shutdown
- add regression coverage for the queued-exit reload path

## Testing
- `pnpm lint`
- `pnpm typecheck`
- `pnpm test`

### DIFF
--- a/slack-bridge/README.md
+++ b/slack-bridge/README.md
@@ -179,6 +179,13 @@ User opens Pinet in Slack sidebar
 
 Messages queue while the agent is busy. When the agent finishes, it automatically drains the inbox and responds.
 
+When running as a **broker**, Slack threads now get explicit operational visibility:
+
+- periodic `⏳` progress updates while a turn is still running
+- per-attempt error updates when an assistant turn errors
+- a final failure summary with retry count and provider/model context
+- automatic pause messaging for terminal provider errors (for example usage/auth failures)
+
 ### Available tools
 
 | Tool                         | Description                                                                       |
@@ -283,6 +290,27 @@ Or set `"runtimeMode": "follower"` in settings (or the legacy `"autoFollow": tru
 | `/pinet-exit <agent>`   | Ask another agent to exit                  |
 | `/pinet-free`           | Mark this agent as idle                    |
 | `/pinet-skin <theme>`   | Change the mesh naming theme (broker only) |
+
+### Broker in-thread control commands (from Slack)
+
+In any broker-owned Slack thread, you can now send:
+
+- `pinet reload`
+- `pinet reload <provider/model>`
+- `pinet reload <provider/model> <thinking>`
+- `/pinet-reload <provider/model> <thinking>` (plain-text alias when your Slack client forwards slash text)
+
+Examples:
+
+- `pinet reload openai/gpt-5.4`
+- `pinet reload openai/gpt-5.4 xhigh`
+
+Behavior:
+
+- aborts the active broker turn (same intent as pressing `Esc` in the TUI)
+- applies model/thinking overrides when runtime model controls are available
+- reloads the broker runtime in place
+- resumes queued inbox work after reload
 
 ### How it works
 
--- a/slack-bridge/agent-event-runtime.test.ts
+++ b/slack-bridge/agent-event-runtime.test.ts
@@ -8,6 +8,8 @@ function createDeps(overrides: Partial<AgentEventRuntimeDeps> = {}) {
   const beforeAgentStart = vi.fn(async (event: { systemPrompt: string }) => ({
     systemPrompt: event.systemPrompt + "\nextra guidance",
   }));
+  const onBrokerTurnMessageEnd = vi.fn(async () => {});
+  const onBrokerTurnAgentEnd = vi.fn(async () => {});
   const onCompletionAgentEnd = vi.fn(async () => {});
   const setDeliverTrackedSlackFollowUpMessage = vi.fn();
 
@@ -19,6 +21,8 @@ function createDeps(overrides: Partial<AgentEventRuntimeDeps> = {}) {
     formatError: (error) => (error instanceof Error ? error.message : String(error)),
     deliverFollowUpMessage,
     beforeAgentStart,
+    onBrokerTurnMessageEnd,
+    onBrokerTurnAgentEnd,
     onCompletionAgentEnd,
     setDeliverTrackedSlackFollowUpMessage,
     ...overrides,
@@ -29,6 +33,8 @@ function createDeps(overrides: Partial<AgentEventRuntimeDeps> = {}) {
     deliverFollowUpMessage,
     requireToolPolicy,
     beforeAgentStart,
+    onBrokerTurnMessageEnd,
+    onBrokerTurnAgentEnd,
     onCompletionAgentEnd,
     setDeliverTrackedSlackFollowUpMessage,
   };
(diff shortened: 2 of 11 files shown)


gugu91:
Thanks for pushing this reload-control lane forward. I checked it against current `main`: some of the original ground has already landed there (for example `/pinet-reload` support and queued-`/exit` coverage), while this branch still carries a much larger unique remainder across broker Slack notices / pause handling / override plumbing. Because of that overlap, the next truthful step looks like a scope refresh first: please restate or restack the narrow remainder you still want reviewed from this branch, then it can come back for focused review.

— Silent Slate Mantis

gugu91:
*Will's reviewer agent Lunar Indigo Beaver*

## Verdict: Block

Solid, thoughtful change overall — `onCommandSettled` is a clean generalization, the pause/visibility flow is sensible, and the queued-exit-reload integration test is a good regression anchor. But there's one multi-user reliability bug worth fixing before merge.

`pnpm typecheck`, `pnpm lint`, and the full test suite (70 files / 1106 tests) all pass locally.

## Warnings (should fix)

**1. Concurrent Slack `pinet reload` requests silently drop overrides and Slack feedback — `slack-bridge/index.ts:964`**

`handleSlackBrokerReloadCommand` unconditionally overwrites `pendingBrokerReloadRequest` on every accepted reload, but `queuePinetRemoteControl` collapses later reloads into a single queued slot (`status: "covered"` or `"queued"`) with `scheduledCommand` still `"reload"`. So if A is running, B queues, and C collapses into B, each of B/C overwrites the pending request. On A's settle, `handlePinetRemoteControlSettled` hits `if (event.nextCommand === "reload") return;` and never posts a completion to A's thread or clears A's pending entry. When the queued reload finally runs, only the last writer's `target`/`command` is honored. Other callers see a `🔄 queued` and then nothing, and their model/thinking overrides are silently dropped.

Fix options: (a) reject a second reload while one is pending with a clear "another reload with overrides is already scheduled", (b) keep `pendingBrokerReloadRequest` as an array and fan out per-target completion/failure even when `nextCommand === "reload"`, or (c) emit a neutral `✅ Reload complete` to prior targets before returning early.

**2. New `pi.on("message_end", …)` / `pi.on("agent_end", …)` sit outside `agentEventRuntime` — `slack-bridge/index.ts:1843-1879`**

Works today, but relies on implicit ordering with handlers registered via `agentEventRuntime.register(pi)` (including `slackToolPolicyRuntime.onAgentEnd` and the completion handler that signals idle). Consider moving the new broker turn-state listeners into a small `broker-turn-visibility` runtime so all agent-event wiring lives in one place, or at minimum add a comment documenting the invariant.

**3. `session_shutdown` resets the new broker state, but `session_start` does not — `slack-bridge/index.ts:1932-1937`**

`session_start` is the canonical reset-to-known-good point. Mirror the `brokerAutoDrainPause` / `pendingBrokerReloadRequest` / `activeBrokerTurnState` / `brokerPauseNotifiedThreads` resets there for defense-in-depth.

**4. Error-classification regexes are broad substring heuristics — `slack-bridge/helpers.ts:308-315`**

`RETRYABLE_ASSISTANT_ERROR_PATTERN` and `TERMINAL_BROKER_PROVIDER_ERROR_PATTERN` drive both retry-hint wording and auto-pause behavior. `500|502|503|504|timeout|terminated` will false-positive on innocuous mentions, and `unauthorized|forbidden|auth(?:entication)?\s+failed` will false-pause on transient 401/403 blips. Where possible, classify off structured fields (`stopReason`, provider error codes, HTTP status) and keep the regex only as fallback.

## Suggestions

- `slack-bridge/index.ts:240-245` — log/assert if `activeBrokerTurnState \!== null` before `startActiveBrokerTurnState` so a future regression is loud.
- `slack-bridge/index.ts:305-311` — `postBrokerThreadNotice` uses `botToken\!`; a defensive early-return with `console.warn` protects against a future settings reload that nulls the token.
- `index.ts` has grown past 1900 lines — extracting broker visibility/reload glue would help maintainability.
- Consider adding tests for: (a) two concurrent Slack reload requests, (b) terminal-error → auto-pause → `pinet reload` → resume drain, (c) `pinet reload` rejection when `/exit` is already scheduled.
- Note: reload-from-Slack is gated by the existing Slack allowlist — consistent with existing controls, but worth documenting that any allow-listed user can now swap broker model/thinking via plain text.

## Checklist

- **Correctness**: warn — concurrent reload path drops overrides and thread feedback
- **Architecture & maintainability**: warn — `index.ts` crowded; new event handlers outside `agentEventRuntime`
- **Security & reliability**: pass — allowlist-gated, abort/shutdown paths correct, notice failures swallowed gracefully
- **Tests & verification**: pass — typecheck, lint, and 1106 tests green; new integration test is well targeted

tmustier:
### Findings

- [P2] Broker turn state survives non-session runtime restarts  
  `slack-bridge/index.ts:1613`  
  `stopPinetRuntime()` tears the runtime down without calling `resetBrokerTurnRuntimeState()`. After a terminal provider pause, a successful `/pinet-start` reload or runtime-mode switch leaves `brokerAutoDrainPause` latched (and any progress timer alive), so the restarted runtime still refuses to auto-drain queued inbox work and can keep sending stale progress notices.  
  Clear the broker turn state whenever a full runtime stop/restart happens outside the special Slack-reload handoff, or add an explicit preserve flag for the one path that really needs it.

- [P2] Reload settlement releases the control queue too early  
  `slack-bridge/pinet-remote-control.ts:82`  
  `remoteControlState` is updated before awaiting `onCommandSettled()`. The new Slack reload settlement does async work (posting Slack updates, applying overrides, resuming inbox work), so another `/reload` or `/exit` can start during that await; the first settle handler then still runs with a stale `nextCommand` snapshot and can resume inbox processing against a new in-flight control action.  
  Keep the current command reserved until `onCommandSettled()` finishes, or make the settle path re-check live queue state before draining/applying completion side effects.

### Verdict

**needs attention** — the new broker reload/visibility flow leaves broker turn state behind across non-session restarts and opens a race where a second control action can start before reload settlement finishes.

tmustier:
### Findings

- [P2] Late-queued reloads can lose their override request state  
  `slack-bridge/index.ts:1018`  
  `handlePinetRemoteControlSettled()` snapshots `nextCommand` and clears/promotes `queuedBrokerReloadRequest` before its awaited Slack notice work finishes. `createPinetRemoteControl()` intentionally keeps the current command reserved until `onCommandSettled` returns, so a reload queued during that await window can still start afterward with `activeBrokerReloadRequest === null`. In that case its completion fan-out is skipped, its model/thinking overrides are never applied, and paused inbox draining may resume too early.  
  Re-read the latest queued command/request after the awaited work, or defer request-state promotion/clearing until the settle callback is about to return.

- [P2] Model override commands are accepted on builds that cannot resolve models  
  `slack-bridge/index.ts:941`  
  `validateBrokerReloadOverrideSupport()` only checks for `pi.setModel`, but `applyBrokerReloadOverrides()` later also requires `ctx.modelRegistry.find`. On builds that expose `setModel` without `modelRegistry`, `pinet reload <provider/model>` still performs a real broker restart and only then reports that runtime model switching is unsupported.  
  Reject model-based reloads before scheduling unless `ctx.modelRegistry.find` is available, or resolve the model before starting the reload.

### Verdict

**needs attention** — two P2 correctness issues remain in the Slack reload path, both around preserving and validating reload override state.

gugu91:

PR #513, opened 2026-04-19
fix(slack-bridge): delegate before deep broker inspection

### LINKED ISSUE #407: broker prompt: delegate before deep source inspection to preserve context
## Summary
The broker currently has strong **do not code** guardrails, but it can still burn context by reading too much source code while trying to diagnose issues itself. We should tighten the broker prompt so it delegates earlier instead of performing deep repo/source investigation.

## Why this matters
Thomas wants the broker to preserve context and behave more like a coordinator than a diagnostician.

Right now the broker instructions already say:
- never implement
- always check `pinet_agents` and delegate work
- use bash for read-only inspection and lightweight GitHub coordination

But they do **not** explicitly say that if diagnosis requires reading a lot of code, the broker should stop and hand that off to a worker.

## Current prompt area
`slack-bridge/helpers.ts` currently includes:

- `When a human asks for work to be done, ALWAYS check \`pinet_agents\` for idle workers and delegate via \`pinet_message\`.`
- `Use bash for read-only inspection and lightweight GitHub coordination ...`

That still leaves room for the broker to do too much source reading before delegating.

## Proposed minimal change
Add a short delegate-first clause to the broker prompt.

For example, change this line:

> When a human asks for work to be done, ALWAYS check `pinet_agents` for idle workers and delegate via `pinet_message`. Pick the agent on the right repo/branch when possible.

To something like:

> When a human asks for work to be done, ALWAYS check `pinet_agents` for idle workers and delegate via `pinet_message` **before doing repo/source inspection beyond quick triage**. Pick the agent on the right repo/branch when possible.

That is a small wording change, but it gives the broker a much clearer stopping point.

## Why this is better
- keeps the broker focused on routing and context management
- reduces wasted context window on the coordination process
- pushes real diagnosis into worker sessions that can spend tokens on code safely
- complements existing issues like `#71` / `#107` instead of replacing them

## Acceptance criteria
- broker prompt explicitly says to delegate before deep repo/source inspection
- broker still allows quick triage / GitHub coordination / status checks
- deep diagnosis is treated as worker work, not broker work

tmustier:
## Summary
- tighten broker prompt guidance so quick triage is allowed but deep repo/source inspection is delegated to workers
- keep broker behavior focused on coordination and preserve broker context
- add prompt-guidance coverage for the new delegate-first rule

## Testing
- `pnpm install`
- `pnpm --filter @gugu910/pi-slack-bridge test -- agent-prompt-guidance.test.ts`

Fixes #407


gugu91:
Narrow read-only pass looks good to me. The cut stays in the small `#407` prompt-guidance/test slice, adds the explicit delegate-before-deep-inspection broker guidance, and does not widen into broader runtime or control-plane changes.

— Rapid Crimson Monkey

PR #515, opened 2026-04-19
fix(slack-bridge): surface workers awaiting task report

### LINKED ISSUE #463: slack-bridge: workers silently drop tasks on transient LLM/tool errors (no report, no blocker notice)
## Summary

When a Pinet worker is running a task (received via `pinet_message`) and hits a transient model-API error — typically `Error: fetch failed` on an Anthropic / OpenAI call — the worker is dropped back to its REPL in `IDLE` state **without sending any report or blocker notice to the broker**. The worker still appears healthy in `pinet_agents` (heartbeat fresh), so the broker's RALPH loop doesn't flag it. The task is silently abandoned.

**In a real 17-worker fanout session today, 5 of 17 workers (≈30%) hit this.** None of them self-reported. All would have silently dropped their tasks if a human hadn't asked the broker to verify.

## Repro

Common trigger: any call that hits the upstream LLM API (or an MCP tool backed by one) during the task turn fails with a transient error. Happens often under burst load from many parallel workers.

In the worker's tmux pane:

```
$ gog calendar list --days 90 --json | jq '...'
jq: 2 compile errors
Command timed out after 30 seconds
Took 30.0s

Error: fetch failed

[worker returns to REPL, status IDLE, heartbeat continues fresh]
```

No `pinet_message` back to the broker. No `/pinet-logs` warning. No status change. Work already done — partial findings sit in the worker's context — but nothing is communicated.

## Impact in today's fleet

- 17 workers dispatched with structured task packets
- 5 workers hit `Error: fetch failed` at various points (LinkedIn MCP call, `gog calendar search`, `jq` invocation, monaco API, shm thread fetch)
- All 5 went silent. Context usage was healthy (25-65%). Heartbeat healthy. They just… stopped.
- Broker only recovered them via manual pane inspection + targeted nudge messages with original packet recap re-embedded
- After nudge, all 5 recovered and delivered useful (often near-complete) reports

So the work wasn't wasted — it was _communicably invisible_.

## Why the RALPH loop doesn't catch it

The current heuristics are:
- Heartbeat freshness → ✅ worker is alive
- Status → ✅ `IDLE` looks fine

But a worker that received a task 10 minutes ago and is currently `IDLE` with zero outbound messages sent is **not** in a healthy state — it's stalled. The broker just can't tell from the current signals.

## Proposed fixes

1. **Surface "task in flight" state.** When a worker consumes a `pinet_message` task, mark it `working_on_task:<message_id>`. Keep that flag until the worker sends an outbound `pinet_message` back. A worker that goes `IDLE` without ever responding should be visible in `pinet_agents` as `idle-stalled: pending reply to msg <N> since <ts>`.

2. **Bake "recover and report on fetch-failure" into the worker system prompt.** The bridge injects a persona prompt already. Add an explicit operating rule: _"If you hit any tool/API error mid-task, recover what you can and send a `pinet_message` back to the sender even if partial — never go silent."_ Today's workers aren't instructed to do this; they're trained to call tools and assume the turn continues.

3. **Auto-nudge on stall detection.** If a worker is `idle-stalled` (per (1)) for > N minutes (e.g., 5), bridge auto-sends a follow-up `pinet_message` to the worker: _"You received task X at <ts> and haven't replied. Please report outcome or blocker."_ — similar to what the broker did manually today.

4. **Log fetch-failure exits.** When the worker's turn terminates on an `Error: fetch failed`, log it to `/pinet-logs` so the broker can see "worker X exited abnormally on task Y at <ts>" — even without active diagnosis.

Fix (2) alone would have prevented today's incident. Fix (1) + (3) would have caught it automatically.

## Related

See companion issue _(silent-delivery claim via `pinet_free`)_ — the other half of "worker stopped communicating but still looks fine." A worker tracking `outbound_count_per_task` and `working_on_task` state would cover both failure modes with one instrumentation.

## Environment

- `@gugu910/pi-slack-bridge@0.1.3`
- `pi` latest (as of 2026-04-16)
- 17 parallel workers via tmux + `/pinet-follow`
- Default bridge worker persona / system-prompt injection

tmustier:
## Summary
- surface workers that have accepted tracked tasks but have not reported back yet
- add control-plane and roster visibility for "awaiting report" workers
- cover the new broker-visible signal with focused regression tests

## Testing
- `pnpm --dir slack-bridge test -- broker/control-plane-canvas.test.ts pinet-control-plane-canvas.test.ts home-tab.test.ts pinet-home-tabs.test.ts`
- `pnpm --dir slack-bridge typecheck`

Fixes #463


gugu91:
Thanks for the contribution here, Thomas — this PR now looks superseded by the merged replacement in PR #525, which closed issue #463. The canonical fix landed there, so I’m leaving a short breadcrumb on this branch too.

— Radiant Hazel Koala

Issue #485, opened 2026-04-18
feat: add broker-launchable follower startup contract without process spawning


gugu91:
## Summary
Follow-up narrow slice from #406: define the startup contract that would let the broker launch a follower session without yet taking on the full process-spawn / tmux / daemon actuation problem.

## Why
Read-only preflight on #406 found that the repo already knows how to:
- choose startup/runtime mode
- manage already-connected agents
- route work across the mesh

What it does **not** yet have is a clean launch contract for a broker-started follower session. That is the smallest mergeable next step.

## Scope
- add a broker-launchable follower startup contract / launch spec
- add a process-local startup override that can feed `resolveSlackBridgeStartupRuntimeMode(...)` without mutating shared settings
- keep the slice mechanical, narrow, and testable
- preserve existing registration/runtime guardrails

## Out of scope
- actually spawning a tmux session / terminal / process
- daemon-owned worker lifecycle
- solving end-to-end broker auto-spin-up in this issue

## Likely touchpoints
- `slack-bridge/runtime-mode.ts`
- startup config parsing/helpers
- `slack-bridge/index.ts`
- focused tests around startup contract resolution

## Related
- #406 broader broker-connected worker startup design

## Deliverable
A narrow startup-contract slice that is implementation-ready now and de-risks later worker actuation work.

gugu91:
Closing per current product priority: this startup-contract slice is not relevant right now. The local artifact can stay parked; reopen or replace later if broker-started workers return to the active roadmap.

PR #516, opened 2026-04-19
fix(slack-bridge): clarify stableId reconnect conflicts

tmustier:
## Summary
- trim the earlier `/pinet-spawn-followers` scope out of this PR
- surface a clearer follower reconnect error when the broker rejects a live `stableId` takeover
- add coverage for the stableId conflict hint shown after reconnect failure

## Testing
- `pnpm lint`
- `pnpm typecheck`
- `pnpm test`

### DIFF
--- a/slack-bridge/broker/client.test.ts
+++ b/slack-bridge/broker/client.test.ts
@@ -13,7 +13,11 @@ import {
   computeReconnectDelay,
 } from "./client.js";
 import type { BrokerConnectOpts } from "./client.js";
-import { RPC_AGENT_NAME_CONFLICT, RPC_METHOD_NOT_FOUND } from "./types.js";
+import {
+  RPC_AGENT_NAME_CONFLICT,
+  RPC_AGENT_STABLE_ID_CONFLICT,
+  RPC_METHOD_NOT_FOUND,
+} from "./types.js";
 
 // ─── Helpers ─────────────────────────────────────────────
 
@@ -24,7 +28,13 @@ interface MockServer {
   received: string[];
   close: () => Promise<void>;
   respondTo: (conn: net.Socket, id: number, result: unknown) => void;
-  respondError: (conn: net.Socket, id: number, code: number, message: string) => void;
+  respondError: (
+    conn: net.Socket,
+    id: number,
+    code: number,
+    message: string,
+    data?: unknown,
+  ) => void;
   connectOpts: BrokerConnectOpts;
 }
 
@@ -67,8 +77,17 @@ function createMockServer(port = 0): Promise<MockServer> {
           const msg = JSON.stringify({ jsonrpc: "2.0", id, result }) + "\n";
           conn.write(msg);
         },
-        respondError: (conn, id, code, message) => {
-          const msg = JSON.stringify({ jsonrpc: "2.0", id, error: { code, message } }) + "\n";
+        respondError: (conn, id, code, message, data) => {
+          const msg =
+            JSON.stringify({
+              jsonrpc: "2.0",
+              id,
+              error: {
+                code,
+                message,
+                ...(data === undefined ? {} : { data }),
+              },
+            }) + "\n";
           conn.write(msg);
         },
       });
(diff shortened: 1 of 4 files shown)


gugu91:
Thanks for pushing on this, Thomas. Quick scope-refresh note from current repo truth: this PR is still carrying the broader broker follower spawn-plan path (`/pinet-spawn-followers`, launcher plan wiring, README flow), while the narrower startup-contract slice that had been the truthful next step was later split out and then closed as not active right now (#485).

If you want to keep this lane moving, could you please refresh the scope first before review — either trim it down to the still-desired narrow startup-contract remainder only, or reframe it against the current roadmap?

— Radiant Hazel Koala

tmustier:
Thanks — I refreshed the scope before asking for more review.

This branch no longer carries the `/pinet-spawn-followers` launcher-plan path or README flow. I force-pushed a narrow re-scope so the PR is now just the stableId reconnect-conflict UX fix:
- clearer follower reconnect error when the broker reports `AGENT_STABLE_ID_CONFLICT`
- focused test coverage for the new hint

Also re-ran `pnpm lint`, `pnpm typecheck`, and `pnpm test` at the repo root after the rewrite.

tmustier:
@gugu91 ^

gugu91 (changes requested):
*Will's reviewer agent Sonar Dolphin*

## Verdict
- **Block**
- The added stableId hint looks nice in isolation, but following the signal through the client stack shows the branch is effectively dead code in production: the broker's real error message doesn't contain the substring being matched, and the client's reconnect path never invokes the terminal handler for stableId conflicts in the first place. The new test passes only because it fabricates an error string that real runtime never produces.

## Findings

### Critical (must fix)

- **`slack-bridge/index.ts:1210` — substring check can never match a real broker error.**
  The broker's `RPC_AGENT_STABLE_ID_CONFLICT` response sets `error.message` to `Agent stableId "..." is already active on another live connection. Wait for that agent to disconnect before retrying.` (see `slack-bridge/broker/socket-server.ts:563`). The literal `AGENT_STABLE_ID_CONFLICT` only lives on `error.data.code`, not on the message. `msg(err)` returns `err.message`, so `.includes("AGENT_STABLE_ID_CONFLICT")` is always false in production. The new hint never surfaces.
  *Fix:* detect via the typed error fields — check `err.code === RPC_AGENT_STABLE_ID_CONFLICT` or `(err.data as { code?: string })?.code === "AGENT_STABLE_ID_CONFLICT"` (mirroring `isRpcAgentNameConflictError` in `broker/client.ts`).

- **`slack-bridge/broker/client.ts:699` — terminal reconnect handler is only fired for `AGENT_NAME_CONFLICT`.**
  Even if the message check above were fixed, `handleTerminalReconnectFailure` is only reachable via `client.onReconnectFailed(...)`, and `reconnectOnce` only calls `reconnectFailedHandler` inside `if (isRpcAgentNameConflictError(reconnectError))`. On a live-stableId conflict the client keeps retrying with exponential backoff and never bubbles a terminal failure up to the UI. This PR is therefore a UX-only change whose UX is unreachable.
  *Fix:* add an `isRpcAgentStableIdConflictError` helper and extend the terminal branch in `reconnectOnce`. Or, if keeping retries for stableId conflicts is intentional, drop the hint change entirely — it would be noise.

- **`slack-bridge/index.test.ts:3007` — test fabricates a message suffix (`… code=AGENT_STABLE_ID_CONFLICT`) that no code in the repo emits.**
  I grepped the tree; no production path appends `code=<X>` to `Error.message`. The test green-washes the substring check above: it passes only because the test string was hand-crafted to match the literal the production code searches for. CI stays green, but the real broker→client→notify pipeline is broken.
  *Fix:* build the test error as a real `BrokerRpcRequestError` (message from `socket-server.ts`, `code: RPC_AGENT_STABLE_ID_CONFLICT`, `data: { code: "AGENT_STABLE_ID_CONFLICT", ... }`) and assert on the fields the production code actually consults.

### Warnings (should fix)

- **`slack-bridge/index.ts:1213-1215` — composed message contradicts itself for stableId conflicts.**
  The outer text still says "Update slack-bridge.agentName/agentEmoji or PI_NICKNAME, or clear the explicit identity request" — but stableId is derived from the session file path, not name/emoji/nickname. Users will fiddle with the wrong knobs.
  *Fix:* branch the whole message on conflict type; for stableId, drop the agentName/emoji/nickname guidance entirely.

- **`slack-bridge/index.ts:1211` — "wait about 5s" is tied to `REQUEST_TIMEOUT_MS`, the wrong constant.**
  `REQUEST_TIMEOUT_MS` is the per-RPC-request deadline, not the socket-cleanup time. The conflict clears when the old connection's `close` event fires in `findLiveStableIdConflict`; if the other worker hangs, the relevant bound is `DEFAULT_HEARTBEAT_TIMEOUT_MS = 15_000`, not 5s.
  *Fix:* anchor to the heartbeat timeout, or soften the wording ("wait a few seconds for the other worker to disconnect").

### Suggestions (consider)

- The ~110-line test scaffolding duplicates the earlier name-conflict reconnect test almost verbatim; a small harness helper would reduce drift.
- Once the real stableId terminal path is wired, add a `client.test.ts` assertion that `performRegister` rejection with `RPC_AGENT_STABLE_ID_CONFLICT` triggers `reconnectFailedHandler` (mirroring the name-conflict coverage).

## Checklist
- Correctness: **fail** — hint is unreachable via the real client + broker error shape.
- Architecture & maintainability: **pass with caveats** — substring matching on error messages instead of typed codes is a recurring smell.
- Security & reliability: **pass** — no regression; related #510 redaction not disturbed.
- Tests & verification: **fail** — new test passes only against a fabricated error shape.

## Summary
- UX intent is reasonable, but the implementation pattern-matches on a string the broker never emits and hooks into a terminal path the client never takes for stableId conflicts.
- The new test passes only because the fabricated error string is crafted to satisfy the buggy production check — a false-green.
- Direction: detect by RPC error code, extend `reconnectOnce` to treat `RPC_AGENT_STABLE_ID_CONFLICT` as terminal (or explicitly keep retrying and drop this UX), and branch the notification so stableId remediation isn't mixed with name/emoji/nickname advice.
- Blocking on the correctness gap; happy to re-review once the signal flows end-to-end and the test asserts the real error shape.

tmustier:
LOL

tmustier:
🦥 `Nimbus Olive Sloth` reporting from `~/projects/extensions-pr-516@Thomass-MacBook-Pro.local`

### Findings

- [P2] Retryable stableId conflicts now abort auto-reconnect  
  `slack-bridge/broker/client.ts:715`  
  `RPC_AGENT_STABLE_ID_CONFLICT` is the broker’s “live holder still exists” case, and the broker marks it `retryable: true`; once that holder disconnects, the same stableId can register successfully. Treating it like a terminal name conflict turns a transient reconnect race into a permanent disconnect that now requires a manual `/pinet-follow`.  
  What should change: keep retrying on stableId conflicts, or only stop reconnecting for a distinct non-retryable stableId error.

- [P2] The new 15s stableId hint promises recovery the broker does not perform  
  `slack-bridge/index.ts:1208`  
  The added guidance says to “wait about 15s for the old socket to clear”, but the broker’s conflict check is based on another open socket still being present, and heartbeat pruning only marks DB rows stale — it does not close that socket. In practice, waiting 15s will still leave the same conflict unless the other worker actually disconnects.  
  What should change: tell users to disconnect/stop the other worker, or change the broker to reap stale sockets before advertising a timeout-based retry.

### Verdict

**needs attention** — the change makes stableId reconnect conflicts terminal even though the broker exposes them as retryable, and the new recovery hint points users to a wait path that does not clear the conflict.

tmustier:
### Findings

- [P2] Retryable stableId conflicts are counted across plain connect failures  
  `slack-bridge/broker/client.ts:696`  
  `stableIdConflictAttempt` survives the `connectSocket()` failure path, so a reconnect session that alternates between transport errors and retryable `AGENT_STABLE_ID_CONFLICT` responses can still hit the 3-conflict terminal stop even though the conflicts were not consecutive. That makes the new retry path give up earlier than intended for transient broker restarts or network flaps.  
  Reset `stableIdConflictAttempt` whenever a reconnect attempt fails before `register`, or otherwise scope the cap to consecutive stableId-conflict responses only.

### Verdict

**needs attention** — retryable stableId reconnects can still stop prematurely because the new conflict budget is not reset on ordinary connection failures.

tmustier:
Addressed the review points end-to-end:

- detect stableId conflicts via typed RPC fields instead of message substring matching
- use the real broker error shape in tests
- branch stableId guidance away from name/emoji advice
- keep transient retryable stableId conflicts retrying, but stop after repeated conflicts so the terminal UX is reachable without turning a handoff race into an immediate hard stop
- reset the stableId conflict streak when reconnect fails before re-registering

Validation: `pnpm lint`, `pnpm typecheck`, `pnpm test`, plus GitHub `quality` check are green.

gugu91:

PR #518, opened 2026-04-19
feat(transport): add normalized outbound content

### LINKED ISSUE #403: feat: add transport-aware rendering/parsing across Slack, Markdown, and iMessage
## Summary
Pinet currently treats cross-transport message content as a plain `text: string`, which makes Slack `mrkdwn`, Markdown, and iMessage/plain-text rendering drift from each other. We need a transport-aware content model plus per-transport render/parse adapters.

## Why this matters
Thomas specifically wants text rendering/parsing optimized per platform instead of hoping one string works everywhere.

Today the stack is effectively:
- `transport-core/index.ts` → `InboundMessage.text` / `OutboundMessage.text`
- `slack-bridge/broker/adapters/slack.ts` → posts `msg.text` directly to `chat.postMessage`
- `imessage-bridge/send.ts` → sends plain `text` via AppleScript

That means we lose the distinction between:
- Markdown semantics
- Slack `mrkdwn` semantics
- iMessage plain text / rich text constraints
- future transports with their own formatting rules

## Current pain
Examples this architecture cannot represent cleanly:
- a Markdown list that should render as proper Slack bullets but plain readable text in iMessage
- links/code/emphasis that should be downgraded differently per transport
- incoming Slack rich text / blocks that should be normalized before forwarding to another transport
- agent replies that currently inherit Slack-flavored formatting when relayed elsewhere

## Proposed direction
Introduce a normalized message-content layer, for example:
- a transport-neutral content AST / structured rich-text model
- parser adapters: Slack inbound → normalized content, Markdown input → normalized content
- renderer adapters: normalized content → Slack `mrkdwn` / blocks, iMessage-safe plain text, plain Markdown, etc.

Minimum useful scope:
1. define the normalized content shape
2. render Slack + iMessage from it
3. parse current Slack inbound block/file context into that shape
4. keep backward-compatible plain-text fallback where needed

## Related existing work checked
I searched existing open/closed issues and did not find an open issue that directly tracks transport-aware rendering/parsing.

Related but not the same:
- #373 — iMessage transport wiring
- #397 — extracted `imessage_send` tool registration
- #399 — thread confirmation refactor

## Acceptance criteria
- transport core no longer assumes a single presentation string is the canonical content model
- Slack outbound formatting is transport-aware instead of plain pass-through
- iMessage outbound formatting is transport-aware instead of plain pass-through
- inbound Slack formatting/context can be normalized before routing across transports
- documented rules for Markdown vs Slack vs iMessage rendering

tmustier:
## Summary
- add a normalized outbound content shape with canonical plain fallback plus optional Markdown and Slack-native renderings
- thread transport-aware outbound content through broker transport send paths, Slack rendering, and iMessage output
- document fallback/rendering rules while staying backward-compatible with existing `text`/`blocks` callers

## Testing
- `pnpm vitest run slack-bridge/broker/message-send.test.ts slack-bridge/broker/client.test.ts slack-bridge/broker/adapters/slack.test.ts slack-bridge/imessage-tools.test.ts imessage-bridge/adapter.test.ts`
- `pnpm typecheck --filter @gugu910/pi-slack-bridge --filter @gugu910/pi-imessage-bridge --filter @gugu910/pi-transport-core`

Fixes #403


gugu91:
Thanks for carrying this transport-content lane forward. The current gate looks mechanical rather than conceptual: PR #518 is `OPEN / DIRTY` and `mergeable: CONFLICTING`, while issue `#403` is still open and there is no obvious merged replacement for this feature. Please rebase / resolve conflicts first; once the branch is clean again, it can come back for scoped review.

— Silent Slate Mantis

tmustier:
@gugu91 Conflicts are resolved now. I merged the latest `main` into `feat/403-transport-aware-content`, carried the transport-aware outbound `content` / `blocks` changes through the new `broker-core` extraction, and re-ran `pnpm lint`, `pnpm typecheck`, and `pnpm test` successfully. PR #518 is now `MERGEABLE / CLEAN` and ready for scoped review.

gugu91:
*Will's reviewer agent Shell Sleuth*

## Verdict: Block (small polish pass)

Functionally solid — tests (147/147), typecheck, and lint all clean; `content` threads cleanly through broker → Slack/iMessage adapters and stays backward-compatible with existing `text`/`blocks` callers. Two medium concerns worth fixing before this becomes the base for follow-up PRs.

### Warnings (should fix)

1. **Duplicated `NormalizedMessageContent` shape.** `transport-core/index.ts:13` exports the canonical type, but it's redeclared inline in `broker-core/message-send.ts:34-38`, `slack-bridge/broker/client.ts:368-372`, and (as a narrowed variant) in `slack-bridge/imessage-tools.ts:15-18`. Drift risk is real once this type is extended (more renderings, per-transport variants). Fix: import `NormalizedMessageContent` from `@gugu910/pi-transport-core` — re-exporting via `broker-core/types.ts` alongside `OutboundMessage` keeps it ergonomic.

2. **iMessage adapter prefers `markdown` over `text`.** `resolveIMessageBody` in `imessage-bridge/send.ts:85-95` (and `adapter.ts:67-72`) picks `markdown` whenever it's a non-empty string. iMessage has no markdown renderer, so the moment a caller supplies real markdown the recipient literally sees `**bold**`. Today all callers pass `markdown === text` (see `slack-bridge/imessage-tools.ts:88-91`), so it's latent — but this PR ships the contract future callers will build on. Options: (a) flip the precedence (prefer `text`, fall back to `markdown` only when `text` is missing) — matches the "canonical plain fallback" framing in the README; or (b) rename the field to something like `rich`/`formatted` and state in the README that the value must be iMessage-safe.

### Suggestions (consider)
- `slack-bridge/broker/socket-server.ts:764-782` — when `content.text` is `""`, the whole `content` (including `slackBlocks`) is silently dropped; caller gets no feedback and may be surprised when Slack renders only the plain `body`. Return `RPC_INVALID_PARAMS` or at least log a warning.
- `broker-core/message-send.ts:88-101` — `outbound.text` uses `content?.text ?? body` (after trim), but `db.insertMessage(..., body, ...)` persists raw `input.body`. If the two ever diverge, stored history ≠ delivered message. Either persist `content?.text ?? body` or assert they match / document "body wins for persistence" explicitly in `transport-core/README.md`.
- No direct unit test for `handleMessageSend`'s content validator (empty-text drop, non-object `slackBlocks` stripping). Small table-driven test would lock down the wire-format contract.

### Checklist
- Correctness: **pass** — only live risk (iMessage + real markdown) needs a caller that doesn't exist yet.
- Architecture & maintainability: **fail** — inline type duplication on a foundational contract.
- Security & reliability: **pass** — no new auth/secret paths; block arrays filtered to object entries before forward.
- Tests & verification: **pass** — 147 tests green, typecheck clean, lint clean.

### Commands run in review
- `gh pr view 518`, `gh pr diff 518`
- `pnpm vitest run slack-bridge/broker/message-send.test.ts slack-bridge/broker/client.test.ts slack-bridge/broker/adapters/slack.test.ts slack-bridge/imessage-tools.test.ts imessage-bridge/adapter.test.ts` — 147/147 pass
- `pnpm typecheck --filter @gugu910/pi-slack-bridge --filter @gugu910/pi-imessage-bridge --filter @gugu910/pi-transport-core` — clean
- `pnpm -r lint` — clean

tmustier:
### Findings

- [P3] Empty `content.slackBlocks` suppresses fallback `blocks`  
  `slack-bridge/broker/adapters/slack.ts:150`  
  The new `msg.content?.slackBlocks ?? msg.blocks` precedence treats `[]` as “present”, so a caller that builds `slackBlocks` programmatically can accidentally override a non-empty legacy `blocks` payload with an empty array. Because the next spread also treats `[]` as truthy, the adapter sends an empty block payload instead of the intended blocks.  
  Only prefer `content.slackBlocks` when it has entries; otherwise fall back to `msg.blocks` and omit empty arrays.

### Verdict

**correct** — no P0-P2 issues found; only a low-priority edge case in the new Slack block fallback logic.

gugu91:

Issue #404, opened 2026-04-15
bug: surface broker restart/reload failures with structured operator-visible diagnostics


tmustier:
## Summary
We have many historical fixes around broker reload/restart behavior, but the current runtime still hides or under-surfaces restart/error states. We should open a focused issue for restart/error observability and failure classification.

## Why this matters
Thomas asked to check whether an active issue already exists for broker restarts/errors. I found several *closed* fixes, but no open issue that directly tracks the current visibility/diagnostics gap.

## Existing related issues checked
Historical fixes include:
- #371 — `/pinet-start` reload active runtime
- #375 — busy-session safety on reload
- #279 — Slack tools failing with `shutdown in progress` during reloads
- #78 / #56 — follower reconnect after broker restart
- #368 / #369 — stale lock/socket aftermath after broker death

These are helpful history, but they do not cover the current observability problem end-to-end.

## Current evidence in code
A few places still make restart/error diagnosis harder than it should be:
- `slack-bridge/follower-runtime.ts` swallows poll failures with `catch { /* broker may be restarting */ }`
- follower reconnect registration refresh failures are logged to console but not surfaced in a structured/operator-visible way
- multiple runtime paths distinguish `ok | reconnecting | error | off`, but there is no single failure taxonomy or operator-facing explanation for why the mesh is degraded

## Problem statement
When the broker is restarting, degraded, or partially recovered, operators/workers need to know:
- is this an expected reload vs a real crash?
- is Slack ingress down, broker socket down, DB/lock state broken, or worker registration stale?
- what action should the operator take next?

Right now too much of that state is implicit, transient, or console-only.

## Proposed direction
Add structured restart/error observability:
- classify runtime failures (`reload`, `crash`, `socket unavailable`, `registration refresh failed`, `shutdown in progress`, etc.)
- surface them consistently in broker/follower status and activity logs
- preserve the last meaningful failure reason instead of swallowing it
- add regression tests for restart/degraded-state UX

## Acceptance criteria
- follower/broker paths no longer silently swallow restart-related failures without preserving reason
- operator-visible status explains the current degraded mode and likely next step
- restart/reload vs crash/failure are distinguishable
- tests cover restart, reconnect, failed registration refresh, and partial recovery paths

gugu91:
Resolved by PR #524.

PR #519, opened 2026-04-19
fix(slack-bridge): preserve degraded restart diagnostics

tmustier:
## Summary
- preserve meaningful follower degraded reasons across disconnect/reconnect instead of swallowing them before operator notices
- distinguish reconnect-after-degraded-state from ordinary reconnects and surface degraded broker status for control-plane canvas and Home tab failures
- wire follower degraded-state persistence into runtime state on top of the existing broker reload visibility stack

## Testing
- `pnpm --filter @gugu910/pi-slack-bridge test -- --runInBand slack-bridge/helpers.test.ts slack-bridge/follower-runtime.test.ts slack-bridge/index.test.ts`

Stacked on #511.
Refs #404

### DIFF
--- a/slack-bridge/README.md
+++ b/slack-bridge/README.md
@@ -179,6 +179,13 @@ User opens Pinet in Slack sidebar
 
 Messages queue while the agent is busy. When the agent finishes, it automatically drains the inbox and responds.
 
+When running as a **broker**, Slack threads now get explicit operational visibility:
+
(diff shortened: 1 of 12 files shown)


gugu91:
Thanks for sending this up, Thomas — this lane now looks superseded by the merged replacement in PR #524 for issue #404. The canonical fix landed there, so I’m leaving a quick pointer here to the merged path.

— Radiant Hazel Koala

PR #536, opened 2026-04-21
feat(slack-bridge): add slack_delete tool

### LINKED ISSUE #535: slack-bridge: add slack_delete tool for removing bot-posted messages
## Problem

The slack-bridge extension exposes a rich set of posting and modification tools (`slack_post_channel`, `slack_send`, `slack_modal_open`, `slack_canvas_update`, etc.) but no way to **delete** a message or thread reply that was posted by the bot.

This surfaces whenever a posted message needs to be recalled — layout bugs, wrong channel, superseded drafts, iteration trails that shouldn't be archived.

### Today's repro
While iterating on the `#market-intel-digest` Slack post layout, the bot posted three top-level messages (two early-iteration versions, one final v3). Neither the bot via pi (no `chat.delete` tool) nor the human author (since messages were posted under the bot identity) can remove the superseded posts without workspace-admin intervention.

## Proposal

Add a `slack_delete` tool to the extension. Should support:

- **Delete by `ts` + `channel`** — the canonical Slack `chat.delete` form
- **Delete entire thread** — convenience, iterate over `conversations.replies` and delete each (with a safety prompt in pi before doing it, mirroring the `slack_confirm_action` pattern for destructive ops)

## Suggested shape

```ts
slack_delete({
  channel: string,   // channel name or ID; uses defaultChannel if omitted
  ts: string,        // message timestamp
  thread: boolean?,  // if true, delete all replies too. default false.
})
```

### Safety

Deletion is destructive and irreversible from the bot's side. Either:
- Require explicit `confirm: true` parameter, or
- Auto-route through `slack_confirm_action` first (matches the existing dangerous-action pattern in the bridge)

My preference: explicit `confirm: true` param for single-message deletes; `slack_confirm_action` for thread deletes.

## Why this matters

- Clean recovery from posting mistakes (layout bugs, wrong channel, typos)
- Support for iterative workflows where a draft gets posted for review then replaced
- Parity with how human Slack users operate — they can delete their own messages; the bot should be able to too

## Scope

Just the one tool. Scoped to messages the bot itself posted (Slack API enforces this: `chat.delete` only works on messages posted by the same bot/user token).

## References

- Slack API: https://api.slack.com/methods/chat.delete
- Today's trigger: iteration of the market-intel daily digest (Nexcade/market-intel), where three superseded top-level posts are now stuck in `#market-intel-digest`.

tmustier:
## Summary
- add a new `slack_delete` tool for deleting bot-posted Slack messages
- support whole-thread deletion by enumerating replies and deleting replies before the root
- mark the tool as write/destructive in guardrails, document it, and cover the behavior with tests

## Testing

Closes #535

### DIFF
--- a/slack-bridge/README.md
+++ b/slack-bridge/README.md
@@ -190,6 +190,7 @@ Messages queue while the agent is busy. When the agent finishes, it automaticall
 | `slack_upload`               | Upload files, snippets, or diffs into Slack                                       |
 | `slack_schedule`             | Schedule a message for later delivery                                             |
 | `slack_post_channel`         | Post to a channel (by name or ID)                                                 |
+| `slack_delete`               | Delete a bot-posted message or an entire thread                                   |
 | `slack_read_channel`         | Read channel history or a thread in a channel                                     |
 | `slack_create_channel`       | Create a new Slack channel                                                        |
 | `slack_project_create`       | Create a project channel + RFC canvas + bot invite in one call                    |
--- a/slack-bridge/guardrails.test.ts
+++ b/slack-bridge/guardrails.test.ts
@@ -95,6 +95,7 @@ describe("isToolBlocked", () => {
   it("classifies Slack mutation tools as write-only", () => {
     expect(WRITE_TOOLS.has("slack_create_channel")).toBe(true);
     expect(WRITE_TOOLS.has("slack_post_channel")).toBe(true);
+    expect(WRITE_TOOLS.has("slack_delete")).toBe(true);
     expect(WRITE_TOOLS.has("slack_upload")).toBe(true);
     expect(WRITE_TOOLS.has("slack_schedule")).toBe(true);
     expect(WRITE_TOOLS.has("slack_pin")).toBe(true);
@@ -107,6 +108,7 @@ describe("isToolBlocked", () => {
     expect(READ_ONLY_TOOLS.has("slack_presence")).toBe(true);
     expect(READ_ONLY_TOOLS.has("slack_create_channel")).toBe(false);
     expect(READ_ONLY_TOOLS.has("slack_post_channel")).toBe(false);
+    expect(READ_ONLY_TOOLS.has("slack_delete")).toBe(false);
     expect(READ_ONLY_TOOLS.has("slack_upload")).toBe(false);
     expect(READ_ONLY_TOOLS.has("slack_schedule")).toBe(false);
     expect(READ_ONLY_TOOLS.has("slack_pin")).toBe(false);
--- a/slack-bridge/guardrails.ts
+++ b/slack-bridge/guardrails.ts
@@ -36,6 +36,7 @@ export const WRITE_TOOLS = new Set([
   "memory_init",
   "slack_create_channel",
   "slack_post_channel",
+  "slack_delete",
   "slack_upload",
   "slack_schedule",
   "slack_pin",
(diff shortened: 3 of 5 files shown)


tmustier:
🐒 `Rapid Crimson Monkey` reporting from `~@Thomass-MacBook-Pro.local`

### Findings

- [P1] Whole-thread delete can leave mixed threads partially deleted  
  `slack-bridge/slack-tools.ts:384`  
  `resolveSlackDeleteTargets()` collects every reply ts in the thread, and `slack_delete` then deletes them one by one. In any thread that contains a human reply or another bot’s message, `chat.delete` will fail partway through with `cant_delete_message` after earlier bot replies have already been removed, leaving the thread in an irreversible half-deleted state.  
  Preflight the thread before deleting anything: verify every message is deletable by this bot, or abort the whole-thread path with a clear error before issuing any `chat.delete` calls.

- [P2] Guardrail confirmations won’t match the required `confirm=true` rerun  
  `slack-bridge/slack-tools.ts:1830`  
  The tool includes `confirm=${params.confirm ?? false}` in the `requireToolPolicy()` action string, but `confirm` is only validated afterwards. Confirmation matching is exact, so an approval for the default `confirm=false` call cannot authorize the actual deletion rerun with `confirm=true`, forcing a second approval in guarded Slack threads.  
  Check `confirm` before calling `requireToolPolicy`, and exclude the `confirm` flag from the confirmation action summary.

### Verdict

**needs attention** — whole-thread deletion can irreversibly partially delete mixed-author threads, and the new confirm flag does not integrate cleanly with Slack confirmation guardrails.

tmustier:
🦊 `Atomic Crimson Fox` reporting from `~@Thomass-MacBook-Pro.local`

### Findings

- [P2] Whole-thread preflight still accepts foreign `pi_agent_msg` replies  
  `slack-bridge/slack-tools.ts:204`  
  `isPiAgentSlackMessage()` treats any reply with `metadata.event_type === "pi_agent_msg"` as owned by the current bot, even when Slack identifies a different author. That means a mixed thread can still pass preflight and then fail partway through `chat.delete`, recreating the partial-delete behavior this change was meant to prevent.  
  Only use metadata as a fallback when no concrete author is present, and match it against the current bot/owner metadata instead of `event_type` alone.

- [P2] Thread-delete ownership checks do not request reply metadata  
  `slack-bridge/slack-tools.ts:357`  
  The new thread-delete validation depends on `message.metadata`, but `fetchSlackThreadMessages()` calls `conversations.replies` without `include_all_metadata: true`. Other metadata-based reply reads in this codebase already set that flag, so valid bot-authored threads can be rejected here when Slack omits metadata from the response.  
  Request reply metadata in this call, or avoid relying on metadata for deletability checks.

### Verdict

**needs attention** — whole-thread delete still has ownership-detection gaps that can either reject valid deletes or allow mixed threads through preflight.

tmustier:
@gugu91

gugu91:
Tight review on the `slack_delete` slice only: this stays nicely pinned to the new delete tool seam. The tool is marked destructive in guardrails, whole-thread deletion resolves replies and deletes replies before the root, and the README/tests stay focused on delete behavior without widening into broader Slack mutation policy. Looks good to merge from my side.

— Radiant Hazel Koala

gugu91:
Tight review on the `slack_delete` slice only: this stays nicely pinned to the new delete tool seam. The tool is marked destructive in guardrails, whole-thread deletion resolves replies and deletes replies before the root, and the README/tests stay focused on delete behavior without widening into broader Slack mutation policy. Looks good to merge from my side.

— Radiant Hazel Koala

PR #565, opened 2026-04-23
feat: add compatibility scope carriers to runtime metadata

gugu91:
## Summary
- add first-class optional workspace/install + instance scope carriers to transport-neutral message contracts
- thread compatibility-first scope metadata through broker-core and slack-bridge runtime surfaces without changing live behavior
- keep unknown or empty Slack `teamId` unknown instead of inventing a fake workspace id
- add focused tests/docs for compatibility-mode scope propagation

Closes #546.

## Notes
- carrier-only / compatibility-first slice
- no enforcement or deny-path changes
- no multi-install runtime orchestration
- no operator/admin semantics drift
- no schema migration churn beyond reusing existing broker message metadata JSON

## Testing
- `pnpm --filter @gugu910/pi-transport-core lint`
- `pnpm --filter @gugu910/pi-transport-core typecheck`
- `pnpm --filter @gugu910/pi-transport-core test`
- `pnpm --filter @gugu910/pi-broker-core lint`
- `pnpm --filter @gugu910/pi-broker-core typecheck`
- `pnpm --filter @gugu910/pi-broker-core test`
- `pnpm --filter @gugu910/pi-slack-bridge lint`
- `pnpm --filter @gugu910/pi-slack-bridge typecheck`
- `pnpm --filter @gugu910/pi-slack-bridge test`

PR #567, opened 2026-04-23
feat: add slack install topology compatibility plumbing

gugu91:
## Summary
- define explicit Slack install/workspace topology settings with a compatibility-first default install model
- project the selected default install into the current singleton runtime without adding multi-install orchestration
- preserve install/workspace scope in runtime metadata, thread scope plumbing, and documentation/tests

## Testing
- git push (pre-push ran repo typecheck + test)

### DIFF
--- a/slack-bridge/README.md
+++ b/slack-bridge/README.md
@@ -151,9 +151,11 @@ Slack access is now **default-deny** unless you configure one of these explicitl
 | `appToken`                     | **yes**  | App-Level Token for Socket Mode (`xapp-...`)                                                                       |
 | `allowedUsers`                 | no       | Slack user IDs that can interact; when unset, access is denied unless `allowAllWorkspaceUsers` is true             |
 | `allowAllWorkspaceUsers`       | no       | Explicit opt-in for workspace-wide Slack access when you do not want a user allowlist                              |
-| `defaultChannel`               | no       | Default channel for the `slack` dispatcher `post_channel` action                                                   |
-| `logChannel`                   | no       | Channel for broker activity logs                                                                                   |
-| `logLevel`                     | no       | `"errors"`, `"actions"` (default), or `"verbose"`                                                                  |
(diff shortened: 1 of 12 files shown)


will-jack-and-jilll (commented):

## Verdict
- **Block**
- The happy-path plumbing looks solid, but this slice is not merge-safe for a stacked multi-workspace rollout because explicit topology can silently advertise one install/workspace while still running with fallback legacy tokens/settings from another.

## Top findings
- `slack-bridge/helpers.ts:355-448` — explicit installs silently fall back to top-level/env bot/app tokens and channels when the selected install omits them. That lets `defaultInstallId`/`scope` say `workspace-b` while the runtime still connects with legacy workspace-a credentials. In the next stacked slice (#568), install/workspace scope becomes an authorization boundary, so this drift can deny valid traffic or authorize the wrong boundary. Fail fast when `installs` is present and the selected install is missing required singleton fields, or make the resolved scope derive from the actual credentials in use instead of the declared install.
- `slack-bridge/helpers.ts:220-225,372-375` — normalized install IDs are never validated for uniqueness, and an unknown `defaultInstallId` silently falls back to `installs[0]`. In a stacked chain, later slices treat `installId` as a routing/security identity, so typoing the default or duplicating an install ID can silently target the wrong workspace. Validate duplicates + unknown defaults and refuse startup.
- `slack-bridge/helpers.ts:205-210` — `buildSlackThreadScope` overwrites `workspaceId` with live `teamId` but keeps the old `installId`, which can create impossible scope pairs if the connected token and declared install drift. Reject or surface the mismatch instead of merging mixed identifiers.

## Verification
- `pnpm exec vitest run --configLoader runner helpers.test.ts runtime-agent-context.test.ts single-player-runtime.test.ts broker/adapters/slack.test.ts`

## Stack safety
- Safe as a concept for the `#565 -> #567 -> #568 -> #570` chain only after topology validation is tightened. Right now this PR can mint incorrect install metadata that the next PRs will trust as policy/routing input.

gugu91:

PR #598, opened 2026-04-25
fix(slack-bridge): cap retryable stableId reconnect conflicts

gugu91:
## Summary
- retry broker reconnects for retryable `AGENT_STABLE_ID_CONFLICT` responses, but stop after 3 consecutive conflicts
- reset the stableId conflict streak on successful reconnects, ordinary reconnect failures, manual connect/disconnect, and name-conflict terminal paths
- surface stableId-specific terminal guidance that tells users to stop/disconnect the other worker instead of changing agent name/emoji

## Context
- Respin of the useful, bounded #516 behavior after @gugu91 approved the policy: retry retryable stableId conflicts up to 3 consecutive times, then stop with explicit guidance.
- Related salvage/audit tracker: #577. I did not find a more specific open behavior issue for this exact retry policy.

## Testing
- `pnpm install --frozen-lockfile`
- `pnpm --filter @gugu910/pi-slack-bridge exec vitest run broker/client.test.ts index.test.ts`
- `pnpm --filter @gugu910/pi-slack-bridge typecheck`
- `pnpm --filter @gugu910/pi-slack-bridge lint`
- `pnpm --filter @gugu910/pi-slack-bridge test`
- push pre-push hook (`turbo run typecheck`, `turbo run test`)

## Status
READY FOR HUMAN REVIEW. No merge performed.

PR #622, opened 2026-04-27
fix(slack-bridge): preserve degraded restart diagnostics

### LINKED ISSUE #404: bug: surface broker restart/reload failures with structured operator-visible diagnostics

tmustier:
## Summary
Follow-up to the closed #404 thread.

This PR keeps the already-stacked broker reload work intact and adds a narrow operator-facing diagnostics slice that was still missing:
- preserve the last meaningful follower degraded reason across disconnect/reconnect
- surface degraded follower reconnect context in the reconnect/disconnect notices
- surface broker control-plane/Home-tab degraded state more clearly in `/pinet-status`
- add focused tests for the degraded restart/reconnect UX

## Why this is separate from #404
Issue #404 is already closed, but this branch contains a distinct follow-up diagnostic slice that was not covered by the existing broker reload/reconnect PRs.

## Non-overlap with PR #598
I checked this branch against PR #598 (`cap retryable stableId reconnect conflicts`). PR #598 focuses on retryable stableId conflict handling during follower reconnect. This branch instead focuses on preserving and surfacing degraded restart/reconnect reasons plus operator-visible degraded status messaging.

## Validation
- `pnpm prepush`

### DIFF
--- a/slack-bridge/README.md
+++ b/slack-bridge/README.md
@@ -179,6 +179,13 @@ User opens Pinet in Slack sidebar
 
 Messages queue while the agent is busy. When the agent finishes, it automatically drains the inbox and responds.
 
+When running as a **broker**, Slack threads now get explicit operational visibility:
+
+- periodic `⏳` progress updates while a turn is still running
+- per-attempt error updates when an assistant turn errors
+- a final failure summary with retry count and provider/model context
+- automatic pause messaging for terminal provider errors (for example usage/auth failures)
+
 ### Available tools
 
 | Tool                         | Description                                                                       |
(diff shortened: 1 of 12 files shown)


gugu91:

PR #624, opened 2026-04-27
feat(slack-bridge): add broker follower spawn plan

### LINKED ISSUE #406: feat: let the broker spin up broker-connected workers without manual tmux/session choreography
## Summary
The broker is correctly forbidden from spawning local subagents, but there is no first-class replacement for spinning up broker-connected workers. Today this creates real operator friction: the broker can only say "go start another worker," while humans have to manually open sessions (often via tmux/extra terminals), start Pi, and run `/pinet-follow`.

## Why this matters
Thomas explicitly wants the broker to be able to spin up its own workers more intuitively.

## Current friction observed
The current system intentionally blocks the obvious path:
- `slack-bridge/guardrails.ts` forbids the broker from using `Agent`, `edit`, and `write`
- `slack-bridge/helpers.ts` says local subagents are forbidden because they have no Slack/Pinet connectivity and cannot be monitored
- `slack-bridge/README.md` documents `/pinet-start` and `/pinet-follow` as separate manual commands for separate sessions

So the broker can delegate only to **already-running** followers. If there are no followers, the operator has to manually:
1. open another session/terminal/tmux pane
2. launch Pi in the right repo/context
3. run `/pinet-follow`
4. verify the worker actually registered with the broker

That is safe, but not ergonomic.

## Related existing work checked
Related historical issues, but none solve broker-managed worker bootstrap:
- #64 — broker should be routing-only
- #71 — delegate via Pinet instead of local subagents
- #107 — hard broker guardrails to stop coding
- #156 — keep local subagents out of the Pinet mesh

So the restriction is intentional and correct — what is missing is a safe **broker-managed worker launcher**.

## Proposed direction
Add a first-class worker bootstrap flow, e.g. one of:
- `/pinet-spawn` command
- broker-only tool to launch a local follower process/session
- broker-managed tmux/session launcher
- optional worktree-aware spawn flow for implementation workers

Key requirement: spawned workers must be **real Pinet followers**, not invisible local subagents.

## Requirements for a good solution
- broker can request worker launch without violating coordination-only guardrails
- spawned worker starts in a monitorable session and auto-runs `/pinet-follow`
- repo/branch/worktree/cwd can be specified at spawn time
- broker gets clear success/failure feedback when registration does or does not happen
- spawned worker is distinguishable from ephemeral local subagents
- safe defaults prevent accidental mesh pollution or duplicate workers

## Acceptance criteria
- there is a first-class broker-managed path to spin up a connected worker
- operators no longer need manual tmux/terminal choreography for the common case
- launched workers appear in `pinet_agents` with clear metadata
- failures to launch/register are surfaced back to the broker/operator clearly

tmustier:
## Summary
- add a broker-side `/pinet-spawn-followers [count]` command that prints a launch plan for real Pinet followers
- extract typed spawn-plan generation into a dedicated helper with tests
- document the new broker command and the launcher-plan workflow

## Testing
- pnpm lint
- pnpm typecheck
- pnpm test

Closes #406.

### DIFF
--- a/slack-bridge/broker/client.test.ts
+++ b/slack-bridge/broker/client.test.ts
@@ -13,7 +13,11 @@ import {
   computeReconnectDelay,
 } from "./client.js";
 import type { BrokerConnectOpts } from "./client.js";
-import { RPC_AGENT_NAME_CONFLICT, RPC_METHOD_NOT_FOUND } from "./types.js";
+import {
+  RPC_AGENT_NAME_CONFLICT,
+  RPC_AGENT_STABLE_ID_CONFLICT,
+  RPC_METHOD_NOT_FOUND,
+} from "./types.js";
 
 // ─── Helpers ─────────────────────────────────────────────
 
@@ -24,7 +28,13 @@ interface MockServer {
   received: string[];
   close: () => Promise<void>;
   respondTo: (conn: net.Socket, id: number, result: unknown) => void;
-  respondError: (conn: net.Socket, id: number, code: number, message: string) => void;
+  respondError: (
+    conn: net.Socket,
+    id: number,
+    code: number,
+    message: string,
+    data?: unknown,
+  ) => void;
   connectOpts: BrokerConnectOpts;
 }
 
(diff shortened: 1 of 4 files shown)


tmustier:
Superseded by #625 — original branch name `feat/406-broker-worker-spawn` was already occupied by an unrelated stableId series at `cd84afa`, so this PR does not contain the intended #406 work. Closing in favor of the fresh branch.

PR #630, opened 2026-04-27
fix(repo): include browser-playwright in workspace

### LINKED ISSUE #626: infra: pnpm prepush blocked by unrelated browser-playwright dep/ambient errors
## Summary

The repo-level pre-push hook runs the full monorepo `pnpm prepush` and fails inside the unrelated `browser-playwright` package due to missing/ambient module errors (`@mariozechner/pi-ai`, `@mariozechner/pi-coding-agent`, `playwright`, etc.). This blocks unrelated branches from being pushed even when their own slice is fully validated.

## Discovered while

Pushing a documentation/test slice for #403 follow-up work (branch `fix/403-transport-content-docs-tests`). The slice itself passed:

- `pnpm vitest run slack-bridge/broker/integration.test.ts`
- `pnpm typecheck --filter @gugu910/pi-slack-bridge`

But `git push` was blocked by `browser-playwright` errors that have nothing to do with the change.

## Why this matters

- Unrelated package breakage shouldn't gate every branch push in the monorepo.
- Workers either get blocked or are tempted to rely on `git push --no-verify`, which erodes the policy surface of the hook.

## Suggested fix paths (any one is enough)

1. Ensure `browser-playwright` declares its deps correctly (and they install via `pnpm install`) so `pnpm prepush` is green from a clean checkout.
2. Scope the pre-push hook to only run checks on packages affected by the diff (Turbo's `--filter [HEAD]` style) so unrelated breakage doesn't gate unrelated work.
3. As an interim guard, mark `browser-playwright` as optional in the prepush graph and emit a loud warning instead of a hard fail.

## Acceptance

- A clean monorepo checkout can run `pnpm prepush` to green.
- A worker with a slice that passes its own focused checks can `git push` without `--no-verify`.

## Related

- Original block reported by Atomic Rust Buffalo on the #403 follow-up slice.
- Loosely adjacent to #585 (Turborepo verification) and #588 (extension audit).

tmustier:
## Summary
- include `.pi/extensions/*` in the pnpm workspace so the checked-in `browser-playwright` extension is installed and validated in clean worktrees
- restore green `pnpm prepush` behavior by ensuring the extension's declared deps are actually bootstrapped
- keep the fix minimal: no hook-policy changes, just align workspace/package installation with the existing repo layout

## Reproduction
Before this change, a clean worktree could hit repo pre-push failures from the checked-in `browser-playwright` extension because its package was outside `pnpm-workspace.yaml`, so its declared dependencies were never installed by the repo bootstrap.

## Validation
- `pnpm --dir .pi/extensions/browser-playwright run check`
- `pnpm prepush`

Fixes #626

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -41,6 +41,25 @@ importers:
         specifier: ^4.1.2
         version: 4.1.2(@types/node@22.19.11)(vite@8.0.3(@emnapi/core@1.9.1)(@emnapi/runtime@1.9.1)(@types/node@22.19.11)(jiti@2.6.1)(yaml@2.8.2))
 
+  .pi/extensions/browser-playwright:
+    dependencies:
+      playwright:
+        specifier: ^1.52.0
+        version: 1.59.1
+    devDependencies:
+      "@mariozechner/pi-ai":
+        specifier: ^0.65.2
+        version: 0.65.2(ws@8.20.0)(zod@4.3.6)
+      "@mariozechner/pi-coding-agent":
+        specifier: ^0.65.2
+        version: 0.65.2(ws@8.20.0)(zod@4.3.6)
+      "@types/node":
+        specifier: ^22.13.1
+        version: 22.19.11
+      typescript:
+        specifier: ^5.9.3
+        version: 5.9.3
+
   imessage-bridge:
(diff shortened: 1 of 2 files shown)


gugu91:

PR #711, opened 2026-05-03
feat(pinet): add port lease mechanism

gugu91:
## Summary
- add durable SQLite-backed Pinet port leases with acquire/renew/release/status/list/expire APIs
- wire broker/follower RPC and `pinet action=ports` dispatcher/help surface
- expire stale leases during broker maintenance and document conservative cleanup semantics

## Tests
- pnpm lint
- pnpm typecheck
- pnpm test

## Notes
- Existing worktree branch was originally behind main; push/pre-push checks passed on the branch. If GitHub reports merge conflicts with recent main skin/doc changes, I can rebase/resolve next.


gugu91:
*Will's reviewer agent Lease Llama*

## Verdict
- **Block**
- The port-lease storage itself is close, but the new RPC/dispatcher surface has two authority bugs that let one follower break another follower's lease, and the acquire API does not match the documented requested-port semantics.

## Top findings
- `slack-bridge/broker/socket-server.ts:1146-1254` — the socket layer accepts caller-supplied `ownerAgentId` and also exposes unrestricted `list/status`, so any registered follower can enumerate another agent's lease IDs and then renew/release them by spoofing the owner.
- `slack-bridge/broker/socket-server.ts:1266-1269`, `slack-bridge/pinet-tools.ts:955-956,1341` — `portLease.expire` / `pinet action=ports op=expire` expose a caller-controlled time override. Any agent can pass a future timestamp and force-expire active leases immediately.
- `broker-core/schema.ts:1524-1530`, `slack-bridge/README.md:462` — an explicit `port` is still validated against the default `49152..65535` allocation range, so `op=acquire port=3000` fails unless callers also know to widen `min_port`/`max_port`. That contradicts the documented "requested port" path.

Tests cover DB happy paths and dispatcher acquire/list, but there is no follower/RPC coverage for ownership, expire override, or the requested-port edge case that introduced the blockers above.

gugu91:
*Will's reviewer agent Socket Badger*

Re-review after 754f43d: **Approve**.

The previous blockers look fixed:
- follower RPC now binds acquire/renew/release/status/list to the registered agent, so spoofed `ownerAgentId` values are ignored and other agents cannot inspect or mutate a neighbor's lease
- the follower-facing ports surfaces no longer honor public expire-time or owner overrides
- explicit requested ports outside the default allocation range now work when no explicit range is supplied (for example `port=3000`)
- regression coverage was added for the owner-scoping/override path and the explicit-port path

Verification:
- `pnpm vitest run broker-core/schema.test.ts -t "explicit requested port outside the default allocation range"`
- `pnpm vitest run slack-bridge/broker/integration.test.ts -t "scopes port lease follower RPCs to the registered agent"`
- `pnpm --filter @gugu910/pi-broker-core typecheck && pnpm --filter @gugu910/pi-slack-bridge typecheck`
- `pnpm --filter @gugu910/pi-broker-core lint && pnpm --filter @gugu910/pi-slack-bridge lint`

PR #742, opened 2026-05-11
Fix Pi Git package install

tmustier:
## Summary
- make the monorepo installable via Pi's default Git package path (`npm install --omit=dev`)
- add npm workspaces and npm-compatible local workspace dependency specs
- replace the dev-only Husky prepare failure with a production-safe prepare shim
- add a postinstall build path that emits package dist files needed by workspace package exports

## Why
`pi install git:github.com/gugu91/extensions` currently fails because Pi runs `npm install --omit=dev`; the root `prepare` script calls `husky`, which is dev-only, and the repo also needs dist files built for workspace package exports.

## Verification
- clean clone + overlay patch: `npm install --omit=dev` succeeds
- `node -e "import('./slack-bridge/dist/index.js')..."` succeeds
- `pnpm build`
- `pnpm --filter @gugu910/pi-slack-bridge test`
- pre-push hook ran full `pnpm lint && pnpm typecheck && pnpm test` successfully

### DIFF
--- a/broker-core/package.json
+++ b/broker-core/package.json
@@ -43,6 +43,6 @@
     "test": "vitest run *.test.ts"
   },
   "dependencies": {
-    "@gugu910/pi-transport-core": "workspace:*"
+    "@gugu910/pi-transport-core": "file:../transport-core"
   }
 }
--- a/imessage-bridge/package.json
+++ b/imessage-bridge/package.json
@@ -32,6 +32,6 @@
     "test": "vitest run"
   },
   "dependencies": {
-    "@gugu910/pi-transport-core": "workspace:*"
+    "@gugu910/pi-transport-core": "file:../transport-core"
   }
 }
--- a/package.json
+++ b/package.json
@@ -23,12 +23,14 @@
     "lint": "turbo run lint",
     "typecheck": "turbo run typecheck",
     "build": "turbo run build",
+    "build:packages": "node ./scripts/build-all.mjs",
     "test": "turbo run test",
     "deploy:slack": "node --experimental-strip-types ./slack-bridge/deploy-manifest.ts",
     "check": "pnpm lint && pnpm typecheck && pnpm format:check",
     "prepush": "pnpm lint && pnpm typecheck && pnpm test",
     "lint-staged": "lint-staged",
-    "prepare": "husky"
+    "prepare": "node ./scripts/prepare.mjs",
+    "postinstall": "node ./scripts/build-all.mjs"
   },
   "lint-staged": {
     "*.{ts,tsx,js,mjs,cjs,json,md,yml,yaml}": "prettier --write",
@@ -43,8 +45,22 @@
     "lint-staged": "^16.2.7",
     "prettier": "^3.5.1",
     "turbo": "^2.5.4",
-    "typescript": "^5.7.3",
     "typescript-eslint": "^8.22.0",
     "vitest": "^4.1.2"
+  },
+  "workspaces": [
+    "transport-core",
+    "broker-core",
+    "browser-playwright",
+    "pinet-core",
+    "slack-bridge",
+    "slack-api",
+    "imessage-bridge",
+    "nvim-bridge",
+    "neon-psql",
+    "openai-execution-shaping"
+  ],
+  "dependencies": {
+    "typescript": "^5.7.3"
   }
 }
--- a/pinet-core/package.json
+++ b/pinet-core/package.json
@@ -34,7 +34,7 @@
     "test": "vitest run *.test.ts"
   },
   "dependencies": {
-    "@gugu910/pi-broker-core": "workspace:*",
-    "@gugu910/pi-transport-core": "workspace:*"
+    "@gugu910/pi-broker-core": "file:../broker-core",
+    "@gugu910/pi-transport-core": "file:../transport-core"
   }
 }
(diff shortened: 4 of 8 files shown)


tmustier:
Monorepo critique / better direction after looking at this more carefully:

I agree with the instinct that the repo should get this from pnpm rather than path-style rewrites.

### What is good today

- This is already structurally a pnpm monorepo: `pnpm-workspace.yaml`, `packageManager: pnpm@10.18.1`, `workspace:*` internal deps, and Turbo tasks.
- `workspace:*` is the right expression of intent for internal packages. It guarantees local workspace resolution under pnpm and avoids accidentally pulling similarly named packages from a registry.
- The internal package boundaries are useful (`transport-core`, `broker-core`, `pinet-core`, `slack-bridge`, etc.) and the explicit `exports` maps make the runtime surface clearer.

### What is currently awkward

- The Pi Git install path uses npm semantics (`npm install --omit=dev`) against a pnpm workspace. That is the real impedance mismatch.
- My PR's `file:../...` dependency changes are a compatibility shim for npm, not a better monorepo model. They make npm happy, but they lose the nice `workspace:*` guarantee and encode filesystem paths into package manifests.
- Adding npm `workspaces` duplicates `pnpm-workspace.yaml`, which creates drift risk. If pnpm is the package manager, the workspace graph should ideally come from pnpm only.
- The repo has a hybrid runtime/build shape: Pi extension entries point at TS source (`./slack-bridge/index.ts`), but internal package `exports` point at `dist/*.js`. That means a Git checkout must build some packages before the source entry can import internal packages. This is workable but fragile.
- The custom `scripts/build-package.mjs` / package allowlist is doing monorepo orchestration manually. Turbo/pnpm already know the workspace graph; ideally the repo should not need another hard-coded package order list.
- `turbo.json`'s `build` task does not currently declare `dependsOn: ["^build"]`, so dependency ordering is not encoded in Turbo itself. The manual `build-all.mjs` order compensates for that.

### Preferred long-term shape

1. Keep internal deps as `workspace:*`.
2. Keep the workspace graph in `pnpm-workspace.yaml`; avoid adding npm-only workspace config unless npm is a first-class supported installer.
3. Make Pi's Git package installer package-manager aware:
   - read root `packageManager`
   - if it says pnpm, use `corepack pnpm install` / `pnpm install --prod` or an equivalent production install command
   - do not force npm onto a pnpm workspace
4. Encode build dependency ordering in Turbo, e.g. `build.dependsOn: ["^build"]`, or use `pnpm -r --sort run build` where appropriate.
5. Decide one runtime convention:
   - either Pi loads built JS extension entries (`./slack-bridge/dist/index.js`) and Git install always builds, or
   - Pi/source-loading path avoids internal `dist` exports during local/Git installs.

### Practical conclusion

This PR is a pragmatic fix for today's Pi npm-based Git installer, but if we can change Pi's installer behavior, I would prefer a cleaner follow-up that reverts the `file:../...` manifest changes and lets pnpm resolve `workspace:*` natively.

So yes: "get this from pnpm without paths" is the cleaner monorepo answer. The real fix probably belongs either in Pi's Git package install flow or in package-level installer metadata that tells Pi to use pnpm for this repo.

tmustier:
Hm ok no we are NOT going to patch pi upstream

tmustier:
Docker clean-room repro results:

### Current `main` via Pi Git install

In a fresh `node:24-bookworm-slim` container with `@earendil-works/pi-coding-agent@0.74.0`:

```sh
pi install git:github.com/gugu91/extensions
```

fails exactly as suspected:

```text
> pi-extensions@0.1.4 prepare
> husky

sh: 1: husky: not found
Error: npm install --omit=dev failed with code 127
```

So this is reproducible without local machine state.

### This PR branch via current Pi

```sh
pi install git:github.com/gugu91/extensions@fix/pi-git-install-npm-workspaces
```

succeeds under current Pi, because it is npm-compatible.

### Cleaner pnpm-shaped finding

I also tested the direction you suggested: keep `workspace:*` and let pnpm resolve the workspace graph. A clean clone of `main` with:

```sh
corepack pnpm install --prod
```

still fails on `prepare` because `husky` is dev-only. However, if the repo has only the production-safe `prepare`, production build hook, and `typescript` available as a production install dependency, then:

```sh
corepack pnpm install --prod
```

works with `workspace:*` intact and `slack-bridge/dist/index.js` imports successfully.

So the clean architecture looks like a two-part fix:

1. **Pi upstream:** Git install should respect `packageManager: pnpm@...` and use pnpm instead of forcing npm.
2. **This repo:** keep `workspace:*`; add only the install/build lifecycle fixes needed for a production Git install.

That means the `file:../...` changes in this PR are best viewed as a fallback for current Pi, not the ideal final monorepo shape.

tmustier:
Decision update: we are **not** planning to patch Pi upstream for this.

So the target compatibility contract for this repo is Pi's current Git install behavior:

```sh
npm install --omit=dev
```

That means this repo needs to be installable by npm from a clean Git checkout, even though pnpm remains the preferred development package manager.

Given that constraint, this PR's shape is intentional: it makes the repo tolerate omitted dev deps, makes internal workspace dependencies resolvable by npm, and builds the `dist` files needed by the exported internal packages.

I agree `workspace:*` is cleaner in a pure pnpm monorepo, but without changing Pi's installer, npm compatibility has to live here.

PR #745, opened 2026-05-14
Fix Slack message_replied thread reply ingress

### LINKED ISSUE #744: Slack thread replies can be dropped from broker inbox
## Summary
Human replies in broker/agent-created Slack threads can be visible through Slack history but never reach `slack_inbox` because Slack may deliver threaded replies as hidden `message_replied` updates rather than plain message events.

## Impact
Known Slack threads can lose human replies before persistence/routing, leaving the broker owner with no inbox message.

## Fix in progress
Patch Slack ingress to normalize/hydrate current `message_replied` shapes, preserve existing bot/noise filtering, deduplicate duplicate normal/hidden reply deliveries, and add regression coverage.

gugu91:
## Summary
Fixes #744.

- normalize Slack `message_replied` payloads so known broker/agent-created thread replies route like normal replies
- hydrate body-less known-thread/DM `message_replied` updates via `conversations.replies`, with safe warning logs on hydration failure
- preserve bot/noise filtering, channel type, legacy threaded-DM suppression, and block/file-only message bodies
- deduplicate normal threaded `message` events vs matching hidden `message_replied` updates before hydration/inbox enqueue

## Validation
- `pnpm --filter @gugu910/pi-slack-bridge test -- broker/adapters/slack.test.ts single-player-runtime.test.ts`
- pre-push hook: `pnpm lint && pnpm typecheck && pnpm test`
- local `code-reviewer` subagent final verdict: Approve, no critical/warning findings

## Notes
Slack public docs show `message_replied` as a hidden message subtype carrying thread/root summary fields plus a nested reply object; the fix accepts that current shape and falls back to precise thread hydration when the reply body is omitted.

### DIFF
--- a/slack-bridge/broker/adapters/slack.test.ts
+++ b/slack-bridge/broker/adapters/slack.test.ts
@@ -537,6 +537,190 @@ describe("classifyMessage", () => {
     }
   });
 
+  it("accepts message_replied updates with embedded reply bodies in broker-known threads", () => {
+    const isKnownThread = (ts: string) => ts === "500.600";
+    const evt = {
+      type: "message",
+      subtype: "message_replied",
+      hidden: true,
+      channel: "C1",
+      channel_type: "channel",
+      ts: "500.600",
+      message: {
+        type: "message",
+        user: "U_ROOT",
+        text: "thread root",
+        channel: "C1",
+        thread_ts: "500.600",
+        ts: "500.600",
+        latest_reply: "500.701",
+        replies: [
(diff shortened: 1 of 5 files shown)


gugu91:
Cleanup pass after maintainer triage: closing this stale/dirty PR rather than keeping it as an active merge candidate.

The fix may still be valuable for #744, but this branch is currently dirty against main. If the issue is still active, the safer path is a fresh restack/new PR from current main.

PR #746, opened 2026-05-16
fix(pinet): refresh observed git branch + dirty flag on heartbeat

tmustier:
## Problem

`pinet agents` could keep advertising the launch-time git branch even after a follower's working tree changed. That made agent visibility/routing rely on stale metadata.

## Minimal fix

Refresh only the existing git-related metadata on follower heartbeats:

- Re-probe the worker cwd for the live branch with `git branch --show-current`.
- Re-probe dirty state with `git status --porcelain --untracked-files=normal`.
- Publish the existing `branch` field as the live observed branch, plus:
  - `workdirDirty`
  - `workdirDirtyFileCount`
  - `gitProbeFailed`
  - `gitProbedAt`
- If the branch is detached/unknown, omit `branch` rather than falling back to the cached launch branch.
- If dirty probing fails, do not report the tree as clean; mark `gitProbeFailed` instead.
- Let broker heartbeat accept optional metadata and replace the stored agent metadata in-place without touching identity fields.
- Keep heartbeat liveness protected: the metadata provider has a short timeout and an in-flight guard, so a hung git probe falls back to a plain heartbeat and cannot pile up repeated probes.

Deliberately not included in this smaller version:

- no `observedBranch` / `desiredBranch` split
- no `branchMismatch` model
- no routing score changes
- no broker self-heartbeat refresh
- no large `pinet agents` output redesign

## Tests

- `pnpm prepush`
- Added/updated coverage for:
  - live branch + dirty probing
  - detached/no-branch behavior
  - dirty-probe failure not reported as clean
  - heartbeat metadata timeout/in-flight fallback
  - heartbeat metadata round-trip to broker DB
  - metadata replacement preserving agent identity fields

### DIFF
--- a/broker-core/schema.ts
+++ b/broker-core/schema.ts
@@ -1865,6 +1865,18 @@ export class BrokerDB implements BrokerDBInterface {
     }
   }
 
+  updateAgentMetadata(id: string, metadata: Record<string, unknown> | null): AgentInfo | null {
+    const db = this.getDb();
+    if (!this.getAgentRowById(id)) return null;
+    db.prepare("UPDATE agents SET metadata = ?, last_seen = ? WHERE id = ?").run(
+      metadata ? JSON.stringify(metadata) : null,
+      new Date().toISOString(),
+      id,
+    );
+    const updated = this.getAgentRowById(id);
+    return updated ? rowToAgent(updated) : null;
+  }
+
   updateAgentIdentity(
     id: string,
     identity: { name: string; emoji: string; metadata?: Record<string, unknown> | null },
--- a/slack-bridge/broker/client.test.ts
+++ b/slack-bridge/broker/client.test.ts
@@ -10,6 +10,7 @@ import {
   INITIAL_RECONNECT_DELAY_MS,
   MAX_RECONNECT_DELAY_MS,
   HEARTBEAT_INTERVAL_MS,
+  HEARTBEAT_METADATA_PROVIDER_TIMEOUT_MS,
   computeReconnectDelay,
 } from "./client.js";
 import type { BrokerConnectOpts } from "./client.js";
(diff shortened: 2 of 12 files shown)


tmustier (commented):
Review findings from local pass:

- **[P1] Don’t let metadata refresh block heartbeat liveness** — `slack-bridge/broker/client.ts:886`
  The heartbeat tick waits for the metadata provider before sending any heartbeat. If the git metadata refresh stalls past the broker heartbeat timeout, a live follower sends no heartbeat and can be marked stale/ghost. Race the provider with a short timeout, add an in-flight guard, or send a plain heartbeat independently of metadata refresh.

- **[P2] Surface dirty-probe failures instead of reporting clean** — `slack-bridge/git-metadata.ts:113`
  When `git branch --show-current` succeeds but `git status --porcelain` fails, `probeFailed` remains false while `dirty` is reported as false, so routing/display treat an unknown dirty state as clean. Mark the dynamic probe failed (or add a dirty-probe failure flag) whenever the dirty probe cannot run.

- **[P2] Don’t publish cached branch as a live observed branch after probe failure** — `slack-bridge/runtime-agent-context.ts:420`
  On a failed live probe, this falls back to the cached launch branch as `observedBranch`; routing then gives full branch-match credit even though the current checkout was not observed. Keep the cached value separate from `observedBranch`, or make routing/display treat `gitProbeFailed` as a non-match/warning.

Tests run locally: `pnpm prepush` ✅

tmustier:
Independent review of PR #746:

Verdict: **request changes** (two P2 correctness/robustness issues).

Findings:

1. **P2 — Detached HEAD can keep advertising the cached branch** (`slack-bridge/runtime-agent-context.ts:420`, with `probeGitDynamic` at `slack-bridge/git-metadata.ts:111-118`).
   `git branch --show-current` returns an empty string on detached HEAD, while `git status --porcelain` still succeeds. That yields `dynamic.branch === undefined` and `dynamic.probeFailed === false`, but runtime metadata currently computes `observedBranch = dynamic.branch ?? gitContext.branch`. If a worker launched on `main` later ends up detached, heartbeat metadata will continue publishing the stale cached branch (`main`) rather than “no current branch”, so routing/listing can still treat it as on `main`. The fallback to `gitContext.branch` should only happen when the dynamic probe actually failed, and this needs a detached/no-current-branch test.

2. **P2 — Heartbeat can be blocked indefinitely by the metadata provider** (`slack-bridge/broker/client.ts:881-893`).
   The timer awaits `provider()` before sending any heartbeat RPC. The default provider now performs live git/filesystem work; if it hangs or becomes very slow (filesystem stall, git process hang, etc.), the socket can be healthy but no heartbeat is sent, so the worker may be marked stale/pruned. Provider exceptions are handled, but non-resolution is not. Consider sending a plain heartbeat first, or racing the provider with a short timeout and falling back to a plain heartbeat. Add a test with a non-resolving provider that still expects a heartbeat request.

What looked good: the broker metadata patch path preserves identity fields, metadata-less heartbeats leave existing metadata intact, routing prefers observed branch over desired branch, dirty-tree UI/routing signals are surfaced, and the test coverage is broad for the main happy paths.

Checks run from `.worktrees/review-pr-746`:
- `pnpm install --frozen-lockfile` — passed
- `git diff --check origin/main...HEAD` — passed
- `pnpm --filter @gugu910/pi-slack-bridge test -- git-metadata.test.ts runtime-agent-context.test.ts broker/client.test.ts broker/integration.test.ts helpers.test.ts pinet-tools.test.ts` — passed (74 files, 1312 tests; Vitest ran the slack-bridge suite)
- `pnpm --filter @gugu910/pi-slack-bridge lint` — passed
- `pnpm --filter @gugu910/pi-slack-bridge typecheck` — passed
- `pnpm prepush` — passed via Turbo cache replay

tmustier:
Thanks for the close read @gugu91 — both P2s addressed in 1e81fee.

**1. Detached HEAD fallback (resolved)**
`probeGitDynamic` now distinguishes "probe genuinely failed" from "probe succeeded but no current branch". The shape gains a `detached: boolean` field, and the consumer only falls back to the cached branch when `dynamic.probeFailed === true`. On detached HEAD we publish no `observedBranch` / `branch`, surface `workdirDetached: true`, add a `workdir:detached` capability tag, and (correctly) omit `branchMismatch` since there's no observed branch to compare against. New test: `git-metadata.test.ts > probeGitDynamic > reports detached HEAD when branch is empty but status succeeds`, plus `runtime-agent-context.test.ts > omits observedBranch on detached HEAD instead of falling back to the cached branch`.

**2. Heartbeat starved by a hung metadata provider (resolved)**
Introduced `HEARTBEAT_METADATA_PROVIDER_TIMEOUT_MS = 2500` (strictly less than the 5s heartbeat interval). `runHeartbeatTick` now races the provider against that timeout via a small `runWithTimeout` helper; on timeout (or any provider exception) it falls back to a plain heartbeat so the agent's `lastHeartbeat` stays fresh. New test in `broker/client.test.ts`: heartbeat tick with a never-resolving provider still emits a plain heartbeat after the timeout elapses (uses fake timers via `vi.advanceTimersByTimeAsync` for determinism).

`pnpm prepush` clean. slack-bridge tests: 1315/1315 pass (3 new).

tmustier:
Follow-up re-review after commit `1e81fee`:

Verdict: **approve / no blockers**. The two P2 findings I raised are addressed.

Verified fixes:
- Detached HEAD: `probeGitDynamic` now distinguishes successful no-branch probes with `detached: true`, and `runtime-agent-context` only falls back to the cached branch when `dynamic.probeFailed === true` (`slack-bridge/runtime-agent-context.ts:421-427`). Detached worktrees therefore no longer re-advertise a stale launch branch; metadata/tags surface `workdirDetached` / `workdir:detached` without `branch` / `observedBranch` / `branchMismatch`.
- Heartbeat robustness: heartbeat metadata providers are now bounded by `HEARTBEAT_METADATA_PROVIDER_TIMEOUT_MS = 2500`, and timeout/exception falls back to a plain heartbeat (`slack-bridge/broker/client.ts:124-153`, `925-937`). This closes the hung-provider starvation case.

Checks run from `.worktrees/review-pr-746`:
- `git diff --check a744474..HEAD` — passed
- `pnpm --filter @gugu910/pi-slack-bridge test -- git-metadata.test.ts runtime-agent-context.test.ts broker/client.test.ts` — passed (Vitest ran slack-bridge suite: 74 files / 1315 tests)
- `pnpm --filter @gugu910/pi-slack-bridge lint` — passed
- `pnpm --filter @gugu910/pi-slack-bridge typecheck` — passed

Non-blocking note: detached state is present in runtime metadata/capability tags; the normal agents list mostly shows it by omitting the branch rather than printing an explicit detached warning. That seems acceptable for this fix, but a future small UI polish could render `workdirDetached` explicitly.

PR #757, opened 2026-05-18
Fix Slack thread ownership transfer context

### LINKED ISSUE #756: Bug: Slack thread ownership transfer can leave broker and workers unable to post
## Problem

Slack/Pinet thread ownership transfer is still confusing or broken in recent broker-delegation flows.

Observed today:
- Broker transferred Slack thread `1779139556.450249` to workers via `pinet action=send` with `transfer_thread_id` multiple times.
- Workers still reported they could not post directly because they only had a thread timestamp without enough Slack channel/context, or Slack/Pinet reported that the thread was owned by another agent.
- Broker relay with `slack_send` back to the original thread also failed with: `Thread 1779139556.450249 is already owned by another agent`.
- Human-visible status had to move to side thread `1779139896.435149`, which made coordination confusing.

## Expected behavior

After a broker transfers a Slack thread to a worker, the worker should have enough channel/thread ownership context to post status directly in that Slack thread.

If transfer cannot be completed, or if ownership remains with another agent, the system should expose a clear state and recovery path instead of leaving the broker and worker unable to post in the original thread.

## Repro / investigation notes

A likely repro path:
1. Receive a Slack-originated task in a broker-owned thread.
2. Delegate to worker A with `pinet action=send args.transfer_thread_id=<slack thread_ts>`.
3. Re-delegate or transfer the same Slack thread to worker B.
4. Have the worker attempt direct Slack/Pinet reply to the transferred thread.
5. Have the broker attempt relay to the original Slack thread after ownership has moved.

Check whether the receiving worker gets durable Slack channel + thread context, not only the Slack thread timestamp. Also verify that all transfer-related state changes agree:
- explicit Slack thread owner binding
- Pinet inbox/thread-affinity metadata
- unread row ownership
- any Slack context cache used by `slack_send` / Pinet delivery
- diagnostics and error messages when owner/context is missing

Related prior work: #696, #722, and possibly #744.

## Acceptance criteria

- Explain the current ownership-transfer state machine and where it fails for the observed flow.
- Fix the transfer/reply path if feasible, including regression coverage.
- If a worker cannot receive enough Slack context to post, expose a clear diagnostic/recovery path.
- Preserve first-owner-wins behavior unless the broker explicitly transfers ownership.
- Do not expose Slack tokens, secret values, private prompt contents, or local secret paths in logs, issues, or tests.

gugu91:
## Summary

- append explicit Slack thread transfer context (thread_ts/channel/recovery hint) to broker A2A transfer messages
- store transferred Slack channel/source metadata on `threadOwnershipTransfer`
- surface transferred Slack threads in Pinet inbox pointers with `slack_send` guidance
- hydrate follower local thread state from transfer metadata so transferred workers have channel context available for reply resolution

Fixes #756.

## Investigation notes

The DB ownership transfer path was updating the broker thread owner, but the recipient-visible A2A message stayed opaque: follower polling treats agent messages separately from regular Slack inbox sync, so the worker could receive only a Pinet pointer and generic `reply via pinet action=send` guidance. If the original Slack request had already been read, no unread Slack inbox rows were reassigned, leaving the worker without an obvious Slack reply context/recovery path.

This PR keeps the existing first-owner-wins state machine, but makes explicit broker transfers carry and surface the Slack thread context needed by the new owner.

## Validation

- `pnpm --filter @gugu910/pi-slack-bridge test`
- pre-push: `pnpm lint && pnpm typecheck && pnpm test`

### DIFF
--- a/slack-bridge/follower-runtime.ts
+++ b/slack-bridge/follower-runtime.ts
@@ -16,6 +16,7 @@ import {
   resolvePinetMeshAuth,
   resolveRuntimeAgentIdentity,
   syncFollowerInboxEntries,
+  syncTransferredSlackThreadContexts,
 } from "./helpers.js";
 import {
   type FollowerDeliveryState,
@@ -322,6 +323,18 @@ export function createFollowerRuntime(deps: FollowerRuntimeDeps): FollowerRuntim
           }
 
           if (agentMessages.length > 0) {
+            const transferredThreads = syncTransferredSlackThreadContexts(
+              agentMessages,
+              deps.getThreads(),
+              deps.getAgentOwnerToken(),
+            );
+            if (transferredThreads.threadUpdates.length > 0) {
+              mergeFollowerThreadUpdates(deps.getThreads(), transferredThreads.threadUpdates);
+              if (transferredThreads.changed) {
+                deps.persistState();
+              }
+            }
+
             const pinetPrompt = formatPinetInboxMessages(agentMessages);
             if (deps.deliverFollowUpMessage(pinetPrompt)) {
               markFollowerInboxIdsDelivered(deps.deliveryState, getInboxIds(agentMessages));
--- a/slack-bridge/helpers.test.ts
+++ b/slack-bridge/helpers.test.ts
@@ -60,6 +60,7 @@ import {
   isAgentToAgentEntry,
   partitionFollowerInboxEntries,
   syncBrokerInboxEntries,
+  syncTransferredSlackThreadContexts,
   buildBrokerProtocolGuardrailsPrompt,
   buildWorkerPromptGuidelines,
   buildIdentityReplyGuidelines,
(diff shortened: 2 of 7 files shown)


gugu91:
*Will's reviewer agent Patch Platypus*

## Verdict
- **Block**
- The new transfer context is useful, but the follower-side hydration still preserves a stale local owner for already-cached threads.

Top findings:
- `slack-bridge/helpers.ts:2212` — `syncTransferredSlackThreadContexts()` keeps `existing?.owner` instead of overwriting to the recipient's owner token. For an explicit transfer, that leaves the local cache thinking someone else still owns the thread, so reconnect/reclaim (`slack-bridge/follower-runtime.ts:232`) and `/pinet status` (`slack-bridge/pinet-commands.ts:492`) stay wrong.
- Please add a regression test for the existing-cached-thread/other-owner case.

gugu91:
*Will's reviewer agent Thread Newt*

Re-review after `4a81820` on PR #757: **Approve**.

- Resolved: `slack-bridge/helpers.ts:2212` now overwrites stale cached owner with `agentOwner`, so transferred-thread cache, reclaim, and `/pinet status` converge on the new owner.
- Covered: `slack-bridge/helpers.test.ts:3482` adds the cached-thread/other-owner regression case.

Verification:
- `pnpm --filter @gugu910/pi-slack-bridge test`

No remaining blocking findings from the earlier stale-owner cache review.

PR #758, opened 2026-05-18
Fix Slack thread transfer context sync

### LINKED ISSUE #756: Bug: Slack thread ownership transfer can leave broker and workers unable to post

gugu91:
## Summary
- include Slack source/channel context in explicit Pinet thread-transfer metadata
- sync transferred Slack thread context into follower thread cache before delivering A2A prompts
- keep safe semantics: only broker-explicit Slack transfers with channel context are cached; non-Slack/missing-channel cases remain rejected/ignored

Fixes #756

## Root cause
`transfer_thread_id` updated broker ownership state, but the A2A assignment only carried the Slack thread timestamp. Follower runtime classifies A2A messages separately from normal Slack inbox entries, so workers never persisted the Slack channel/source context needed by `slack_send`/Pinet Slack delivery. Once ownership was explicitly transferred, broker-side relays correctly failed with "already owned by another agent", leaving broker and worker without a usable visible-thread reply path.

## Validation
- `pnpm vitest run slack-bridge/helpers.test.ts slack-bridge/pinet-mesh-ops.test.ts`
- `pnpm lint`
- `pnpm typecheck`
- `pnpm test`
- pre-push hook reran `pnpm lint && pnpm typecheck && pnpm test` successfully

PR author/account: gugu91

### DIFF
--- a/slack-bridge/follower-runtime.ts
+++ b/slack-bridge/follower-runtime.ts
@@ -16,6 +16,7 @@ import {
   resolvePinetMeshAuth,
   resolveRuntimeAgentIdentity,
   syncFollowerInboxEntries,
+  syncFollowerTransferredThreadContext,
 } from "./helpers.js";
 import {
   type FollowerDeliveryState,
@@ -322,6 +323,18 @@ export function createFollowerRuntime(deps: FollowerRuntimeDeps): FollowerRuntim
           }
 
           if (agentMessages.length > 0) {
+            const transferSync = syncFollowerTransferredThreadContext(
+              agentMessages,
+              deps.getThreads(),
+              deps.getAgentOwnerToken(),
+            );
+            if (transferSync.threadUpdates.length > 0) {
+              mergeFollowerThreadUpdates(deps.getThreads(), transferSync.threadUpdates);
+              if (transferSync.changed) {
+                deps.persistState();
+              }
+            }
+
             const pinetPrompt = formatPinetInboxMessages(agentMessages);
             if (deps.deliverFollowUpMessage(pinetPrompt)) {
               markFollowerInboxIdsDelivered(deps.deliveryState, getInboxIds(agentMessages));
--- a/slack-bridge/helpers.test.ts
+++ b/slack-bridge/helpers.test.ts
@@ -83,6 +83,7 @@ import {
   alignAgentIdentityToRole,
   trackBrokerInboundThread,
   syncFollowerInboxEntries,
+  syncFollowerTransferredThreadContext,
   resolveFollowerThreadChannel,
   isDirectMessageChannel,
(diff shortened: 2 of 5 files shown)


gugu91:
Closing as duplicate/superseded by #757.

I compared #757 against this backup PR. #757 covers the material fix here (broker transfer metadata includes Slack source/channel, follower runtime hydrates transferred Slack context before A2A delivery) and adds useful visible recovery guidance / `slack_send` hints that this PR does not. I do not see a material gap that needs to be carried from #758.

Validation for #758 was clean, but #757 should be the coordinated lane for #756.

Issue #772, opened 2026-06-03
Review Pinet neutrality and Slack coupling


gugu91:
Maintainer-approved workstream from Slack thread 1780474385.588819.\n\nGoal: review how neutral Pinet currently is versus how coupled it is to Slack. There has been a goal to separate Pinet from Slack-specific assumptions.\n\nScope notes:\n- Inspect Pinet APIs, broker/worker routing, inbox ownership, identity, reactions, and Slack bridge integration seams.\n- Identify Slack-specific leakage into Pinet core and recommend boundaries.\n- Prefer read-only review unless a small cleanup is clearly safe and approved by issue scope.\n\nAcceptance:\n- Post a concise architecture review and separation recommendations to Slack thread 1780474385.588819.\n- Link any proposed follow-up issues/PRs.

gugu91:
Status update from the current triage pass: keep this umbrella open, but the decoupling track has made concrete progress.

Landed progress:
- #778 closed via #783: broker routing/inbox persistence moved toward transport-neutral records and thread-owner hints.
- #780 closed via #782: runtime composition moved toward injected adapter bindings instead of Slack-first wiring.

Still active:
- #779 remains open with #781 ready for maintainer review/merge: replace Slack-shaped broker RPC boundaries with adapter capabilities.
- #773 publish/release automation remains gated on final package-boundary/version decisions.

Next action: after #781 is decided, update this umbrella with the final architecture conclusion and either close it or split any remaining non-Slack adapter follow-ups into narrower issues.

gugu91:
Post-blitz status update: the current Slack/Pinet decoupling stack is now mostly landed.

Done:
- #778 landed via #783 (transport-neutral routing/inbox persistence).
- #780 landed via #782 (adapter-injected runtime composition).
- #779 landed via #781 (adapter-capability boundary replacing Slack-shaped proxy RPC) and is now closed.

Still open here because this is the umbrella/review issue, not a single implementation slice. Suggested next small slice: audit the remaining Slack-shaped seams after #781, especially names/docs/control surfaces/release packaging assumptions, then either close this umbrella with a final architecture note or file one narrow follow-up for any remaining blocker.

PR #786, opened 2026-06-04
Neutralize broker-core fallback thread sources

gugu91:
## Summary
- default missing-source broker thread upserts/claims to a neutral `external` source instead of Slack
- keep explicit Slack compatibility paths working when callers pass `source: "slack"`
- rename reaction-trigger reclassification audit reason from `slack_reaction_arrow_up` to transport-neutral `reaction_steer`

Refs #772.

## Testing
- `pnpm --filter @gugu910/pi-broker-core test`
- `pnpm --filter @gugu910/pi-broker-core typecheck`
- `pnpm --filter @gugu910/pi-broker-core lint`
- `pnpm --filter @gugu910/pi-slack-bridge test -- broker/router.test.ts broker/message-send.test.ts broker/integration.test.ts`
- push pre-push hook: `pnpm lint && pnpm typecheck && pnpm test`

### DIFF
--- a/broker-core/router.ts
+++ b/broker-core/router.ts
@@ -1,3 +1,4 @@
+import { DEFAULT_EXTERNAL_THREAD_SOURCE } from "./types.js";
 import type { AgentInfo, BrokerDBInterface, InboundMessage, RoutingDecision } from "./types.js";
 
 // ─── Helpers ─────────────────────────────────────────────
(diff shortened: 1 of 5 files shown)


gugu91:
*Will's reviewer agent Patch Badger*

## Verdict
- **Block**
- The neutrality cleanup is directionally right, but it silently changes public thread-claim behavior for callers that omit `source` while still passing a transport channel.

## Top finding
- `broker-core/schema.ts:2200,2328`, `broker-core/router.ts:414`, `slack-bridge/broker/client.ts:507`, `broker-core/message-send.ts:87` — channel-only claims/upserts are no longer Slack-sendable.
  - Before this PR, `thread.claim("t", "C123")` and similar `claimThread(..., channel)` flows created a Slack thread by default.
  - After this PR they create `{ source: "external", channel: "C123" }` unless the caller also passes `source: "slack"`.
  - That breaks follow-up sends that rely on stored thread metadata, because `message.send` resolves `existingThread.source` first and will now fail adapter lookup with `No adapter is registered for transport source "external"`.
  - The new tests do not catch this because `slack-bridge/broker/integration.test.ts:1159` only asserts the stored `channel`, not the `source` or a subsequent send path.
  - Suggested fix: preserve Slack compatibility at the RPC/wrapper layer (or otherwise infer/pass `source: "slack"` for legacy Slack callers), and add a regression test that a channel-only claim still produces a thread that can be sent through the Slack adapter.

PR #788, opened 2026-06-04
fix(pinet): preserve channel-only Slack claims

gugu91:
## Summary
- Preserve legacy Slack compatibility for channel-only broker `thread.claim` RPC calls after #786
- Infer `source: "slack"` at the Slack bridge RPC boundary when callers pass a channel but omit source
- Add regression coverage that channel-only claims remain Slack-sendable via the Slack adapter while source-less claims remain neutral

## Testing
- `pnpm --filter @gugu910/pi-slack-bridge test -- broker/integration.test.ts`
- pre-push hook reran `pnpm lint && pnpm typecheck && pnpm test`

Follow-up to #786 / #772 after review found the post-merge channel-only Slack sendability compatibility risk.

### DIFF
--- a/slack-bridge/broker/integration.test.ts
+++ b/slack-bridge/broker/integration.test.ts
@@ -7,6 +7,7 @@ import { BrokerSocketServer } from "./socket-server.js";
 import { BrokerClient } from "./client.js";
 import { runBrokerMaintenancePass } from "./maintenance.js";
 import { MessageRouter } from "./router.js";
+import type { OutboundMessage } from "./types.js";
 
 // ─── Helpers ─────────────────────────────────────────────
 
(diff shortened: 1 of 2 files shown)


gugu91:
Code-reviewer pass: **Block**.

Critical finding: the hotfix currently fixes fresh channel-only `thread.claim` inserts, but it does not repair rows already persisted by #786 as `{ source: "external", channel: "C..." }`. If one of those rows is reclaimed through the channel-only Slack path, the existing `claimThread` conflict path only updates owner metadata and leaves `source: "external"`, so later `message.send` can still fail with `No adapter is registered for transport source "external"`.

Requested fix:
- When the legacy channel-only Slack path is used, repair existing neutralized rows too, or add a targeted migration/backfill.
- Add a regression test that seeds an `external` + Slack-channel thread, reclaims it channel-only, and verifies `message.send` succeeds.
- Also consider asserting that a truly source-less/no-channel claim stays neutral.

Reviewer validation: focused integration test/typecheck/lint passed, but manual repro exposed the persisted-thread case above.

PR #789, opened 2026-06-04
Preserve channel-only Slack thread claims

gugu91:
## Summary
- preserve legacy Slack compatibility for channel-only `thread.claim` RPC calls by inferring `source: "slack"` at the Slack bridge boundary
- keep broker-core's neutral `external` default for genuinely source-less/non-channel claims
- add regression coverage that a channel-only claim remains sendable through `message.send` via the Slack adapter

Follow-up hotfix for the post-merge #786 compatibility risk noted in review.
Refs #772.

## Testing
- `pnpm --filter @gugu910/pi-slack-bridge test -- broker/integration.test.ts`
- push pre-push hook: `pnpm lint && pnpm typecheck && pnpm test`

## Review
- Local `reviewer` subagent checked the diff and found no blockers/warnings.

### DIFF
--- a/slack-bridge/broker/integration.test.ts
+++ b/slack-bridge/broker/integration.test.ts
@@ -7,6 +7,7 @@ import { BrokerSocketServer } from "./socket-server.js";
 import { BrokerClient } from "./client.js";
 import { runBrokerMaintenancePass } from "./maintenance.js";
 import { MessageRouter } from "./router.js";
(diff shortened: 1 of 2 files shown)


gugu91:
Closing as duplicate: the same channel-only Slack claim hotfix was merged via #788. This branch commit is patch-equivalent to the merged fix, so no separate review path is needed.

Issue #809, opened 2026-06-09
Stop Slack emoji reactions from triggering Pinet actions by default


tmustier:
Thomas requested this after an accidental Slack/Pinet review response caused by a :eyes: reaction trigger in Slack.\n\nDesired behaviour:\n- Slack emoji reactions should not automatically enqueue Pinet action requests or cause the broker/workers to respond.\n- In particular, :eyes: should not map to review/action execution without an explicit text mention/command.\n- Keep normal Slack mentions/DMs/owned-thread replies working.\n- If reaction handling remains useful internally, make it explicit/opt-in and safe by default.\n\nContext:\n- Incident thread: a :eyes: reaction on a Garage PR message was converted into a Pinet review trigger, leading to an unsolicited review posted in Slack.\n- This should be treated as a Slack bridge/admission-policy bug, not a user workflow.\n\nAcceptance:\n- Reaction-trigger events are ignored or downgraded so they do not produce actionable Pinet work by default.\n- Add/update tests covering :eyes: reaction no-op behaviour.\n- Preserve regular message/mention routing.\n- Document any remaining opt-in reaction behaviour if present.

PR #813, opened 2026-06-10
fix(slack-bridge): deny reaction routing in uninvoked threads

### LINKED ISSUE #812: Slack bridge: reaction triggers must not route from uninvoked threads
## Problem
Regression after #809/#810: Slack emoji reaction triggers can still surface as actionable Pinet work in a Slack thread where Pinet/the agent had not previously been invoked.

The rule should be stricter: a bare emoji reaction in an unowned/uninvoked Slack thread should not cause Pinet to enqueue work or post a response, even when the reactor is an authorized user such as Thomas.

## Expected behavior
Reaction-triggered actions should pass the same Slack admission/ownership gate as normal messages:

- If the referenced Slack thread is not already Pinet-owned/known and does not explicitly invoke Pinet/the bot, ignore the reaction for Pinet routing.
- Do not enqueue a Pinet inbox item.
- Do not remember/claim the thread.
- Do not create affinity/ownership side effects.
- Do not post an ACK or review response.

## Allowed behavior
A reaction can route only when at least one of these is true:

- The thread was already explicitly invoked/owned by Pinet.
- The referenced message or thread explicitly mentions Pinet/the bot/agent in a way that passes admission.
- A deliberate opt-in reaction workflow has already established the thread as Pinet-controlled.

## Acceptance ideas
- Test: authorized user reacts with 👀 in a random unowned Slack thread -> no Pinet inbox item and no known-thread side effect.
- Test: authorized user reacts with 👀 in a Pinet-owned Slack thread -> routes according to configured reaction mapping.
- Test: unauthorized user reacts in owned/unowned thread -> ignored.
- Test: reaction handling does not bypass the same `classifyMessage` or admission policy used for normal Slack messages.

This follows the intent of #809/#810; no private Slack content is needed to reproduce.

tmustier:
Closes #812.

## Problem
After #809/#810, a bare 👀 reaction in a Slack thread where Pinet had never been invoked could still produce an actionable Pinet inbox item and a broker post.

**Live root cause (evidence):** the running broker process had been up since Jun 5 and loaded the package dist built *before* #809/#810 landed (checkout updated + dist rebuilt Jun 9 20:59; broker session started 20:58:47). It was running pre-fix code where `eyes → review` was a default mapping with weak gating. **A broker restart is required to pick up the fixed code.**

Independently, this PR closes the remaining *structural* holes in source so reaction routing can never bypass the admission/ownership gate regardless of configuration.

## Changes
- **Deny-by-default reaction authorization**: `SlackAdapter` now requires an explicit `isReactionThreadAuthorized` gate before any opt-in reaction command routes. The previous fallback (`?? !!cachedThread`) let a merely seen/cached thread authorize reactions when no gate was configured. Gate exceptions fail closed without posting visible ❌ reactions into unowned threads.
- **No thread-state side effects from unauthorized traffic**: allowlist checks now run before any thread-state writes in the broker adapter (`onMessage`, `emitInteractiveInbound` — including the durable `rememberKnownThread` write) and the single-player runtime (`onMessage`, `queueInteractiveInboxEvent`). Unauthorized users can no longer mint known-thread/affinity state that would later admit replies or reactions.
- README updated to document the deny-by-default posture.

## Behavior matrix
| Case | Result |
|---|---|
| Authorized user reacts in unowned/uninvoked thread | Ignored: no inbox item, no ACK, no known-thread/ownership side effect |
| Authorized user reacts in Pinet-owned/invoked thread (opt-in mapping + gate) | Routes per configured reaction mapping |
| Unauthorized user reacts (any thread) | Ignored |
| Reaction in cached-but-ungated thread (no authorization callback) | Ignored (new: deny-by-default) |
| Authorization gate throws | Fails closed, no visible error reaction |
| Direct bot mentions / owned-thread replies / DMs | Unchanged |
| Unauthorized message/interactive event | No longer tracks threads or writes durable known-thread state |

## Tests
- `broker/adapters/slack.test.ts`: +3 reaction regression tests (ungated cached thread, explicit gate denial, gate exception), +1 interactive allowlist side-effect test; updated blocked-DM test to assert no thread tracking; existing routing tests now use an explicit gate.
- `single-player-runtime.test.ts`: +1 test (unauthorized message cannot mint thread state that later admits an authorized reaction).
- Full repo: `pnpm lint` ✅, `pnpm typecheck` ✅, `pnpm test` ✅ (10/10 turbo tasks; slack-bridge 79 test files passing).


tmustier:
**Code review (pr-reviewer subagent):** ✅ correct — no P0–P3 findings. Gate denies by default/fails closed; unauthorized reactions dropped before fetch/write side effects; allowlist ordering prevents blocked traffic from minting adapter thread state; owned-thread replies, DMs, mentions, and authorized opt-in reaction flows preserved.

**Validation:** `pnpm lint` ✅ `pnpm typecheck` ✅ `pnpm test` ✅ (10/10 turbo tasks, slack-bridge 79 test files).

**Operational note:** the live incident also requires a broker restart — the running broker process (up since Jun 5) predates #809/#810 and this fix.

PR #815, opened 2026-06-10
fix(slack-bridge): canonicalize guarded delete actions from resolved targets

### LINKED ISSUE #814: Slack delete confirmation registers a different canonical action than enforcement requires
Reported by Thomas (maintainer priority, relayed via broker) after an incident where deleting one mistaken bot reply took multiple confirmation rounds and produced recursive confirmation-prompt spam.

## Problem
1. **Canonical action drift:** `slack_delete` enforces `requireToolPolicy` with an action string built from *raw caller params* (`channel=${params.channel ?? defaultChannel ?? ""}`), while the actual deletion channel is resolved via `resolveSlackTargetChannel` (tracked thread channel → explicit channel → DM → default). The confirmation registered via `confirm_action` must match the enforced string byte-for-byte, so any phrasing difference (omitted channel, #name vs ID, default-channel fallback) makes the approval never match. The delete then fails post-approval and emits a *new* canonical string, prompting a second confirmation.
2. **Recursive prompt spam:** each confirmation prompt is itself a bot message; cleaning up N mistaken/prompt messages requires N approvals (per-message ts in the action string), and deleting the prompts recurses through the same guardrail.
3. A stale mismatched pending request blocks the thread for the 5-minute TTL ("wait for a reply or expiry").

## Proposed fix
- Build the canonical delete action from **resolved** values (resolved channel ID + normalized ts) so confirmation and enforcement always agree regardless of caller phrasing.
- Allow `ts` to enumerate multiple bot-authored messages in one call so a **single explicit approval** covers one batch cleanup (exact messages enumerated in the approved action — no blanket grants).
- Preserve all existing protections: confirm=true, bot-authorship checks for thread deletes, Slack API-level cant_delete_message for non-bot messages.

Source: `slack-bridge/slack-tools.ts` (`summarizeSlackDeleteAction`, `slack_delete`, `slack_confirm_action`), `slack-bridge/thread-confirmations.ts`.

tmustier:
Closes #814. Companion to #813 (same incident family, separate concern).

## Problem
`slack_delete` enforcement built its canonical action string from **raw caller params** (`channel=${params.channel ?? defaultChannel ?? ""}`), while the actual deletion channel resolves via tracked-thread → explicit → DM → default. `confirm_action` registers whatever string the agent passes, and approval matching is byte-exact. So any phrasing drift (omitted channel, #name vs ID, default-channel fallback) meant:
1. approval registered ≠ action enforced → post-approval delete failed,
2. the failure quoted a *new* canonical string → second confirmation prompt,
3. prompts are bot messages whose cleanup recurses through the same guardrail,
4. a stale pending blocked the thread for the 5-minute TTL.

This cost Thomas literal hours when trying to delete a single mistaken bot reply.

## Changes
- **Canonical action from resolved values**: the delete handler resolves the target channel *before* the guardrail check and builds the action string from the resolved channel ID plus a normalized (trimmed/deduped/sorted) ts list. Guardrail errors, confirm_action registrations, and post-approval retries now always agree, so a copied confirmation approves first try regardless of phrasing.
- **Single approval for batch cleanup**: `ts` accepts a comma/whitespace-separated list of bot-posted message timestamps for non-thread deletes. The approved canonical action enumerates the exact timestamps — one explicit approval covers one batch of cleanup messages, no blanket grants. `thread=true` still requires a single root ts.
- **Protections preserved**: `confirm=true` still required; whole-thread deletes still verify pi-agent authorship of every message; Slack's API still rejects deleting other authors' messages with a bot token.

## Tests
- Canonical action equality across omitted/explicit channel phrasings
- Resolved-channel canonical strings (default-channel and tracked-thread paths)
- Batch delete: normalized canonical action, per-ts `chat.delete` calls, result details
- Multi-ts rejection when `thread=true`
- Full repo: `pnpm lint` ✅ `pnpm typecheck` ✅ `pnpm test` ✅ (10/10 turbo tasks)

## Possible follow-up (not in scope)
Auto-deleting the bot's own confirmation prompt after a decision would remove the residual prompt-cleanup step entirely; left out to keep this change focused.

### DIFF
--- a/slack-bridge/slack-tools.test.ts
+++ b/slack-bridge/slack-tools.test.ts
@@ -436,10 +436,13 @@ describe("registerSlackTools", () => {
       confirm: true,
     });
 
+    // The guarded action must carry the RESOLVED channel id so the string in
+    // guardrail errors, confirm_action registrations, and the post-approval
+    // retry always match (#814).
     expect(requireToolPolicy).toHaveBeenCalledWith(
       "slack:delete",
       undefined,
-      "channel=ops-alerts | thread_ts= | ts=123.789 | thread=false",
+      "channel=resolved:ops-alerts | thread_ts= | ts=123.789 | thread=false",
     );
     expect(slack).toHaveBeenCalledWith("chat.delete", "xoxb-reloaded", {
       channel: "resolved:ops-alerts",
@@ -642,7 +645,7 @@ describe("registerSlackTools", () => {
   });
 
   it("uses thread channel resolution for slack_delete", async () => {
-    const { slack, tools, setResolveThreadChannel } = setup();
+    const { slack, tools, setResolveThreadChannel, requireToolPolicy } = setup();
     setResolveThreadChannel(async (threadTs: string | undefined) => {
       expect(threadTs).toBe("123.456");
       return "C-DB";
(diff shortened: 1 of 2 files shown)


tmustier:
**Code review (pr-reviewer subagent):** ✅ correct — no P0–P2 findings. One P3 suggestion: a blocked `slack:delete` now performs a read-only channel resolution before `requireToolPolicy` denies it (resolution must precede the policy check to build the canonical action). Conscious tradeoff: the lookup is read-only metadata the agent can already access via read actions, and avoiding it would require widening the tools deps surface with a separate blocked-tool precheck. Can follow up if maintainers want strict no-API-work-when-blocked semantics.

**Validation:** `pnpm lint` ✅ `pnpm typecheck` ✅ `pnpm test` ✅ (10/10 turbo tasks).

Issue #811, opened 2026-06-10
Slack bridge: avoid treating human-addressed mentions in owned threads as agent-directed work


tmustier:
## Problem
In a Slack thread already owned/known by Pinet, a follow-up message that explicitly mentioned a different human user was still surfaced to the owning Pinet agent as actionable follow-up work.

That is too eager. If the message is clearly addressed to another human participant, Pinet should usually treat it as conversation context, not as an instruction to the owning agent, unless the text separately invokes Pinet/the agent or contains an unambiguous agent-directed request.

## Why this matters
The current owned-thread affinity can over-route human-to-human coordination as agent work. This creates noisy broker routing and risks agents acting on side comments that were not intended for them.

## Desired behavior
Add semantic/structural gating for Slack messages in known/owned threads:

- If the message begins with or prominently contains an explicit mention of another human/non-agent user, down-rank or suppress agent action.
- Still allow routing when the message also explicitly invokes Pinet/the bot/the named agent, or has clear agent-directed imperative content.
- Preserve normal handling for direct mentions of Pinet/bot/agent users.
- Do not create new ownership/known-thread side effects for messages classified as human-directed context.

## Acceptance ideas
- Add tests for an owned Slack thread where a message addressed to a human mention is classified as context/no-action.
- Add tests where a human mention plus explicit bot/agent instruction still routes.
- Add tests ensuring direct bot mentions and regular owned-thread replies remain unaffected.

No private Slack content is needed to reproduce; this is a routing/classification edge case.

PR #856, opened 2026-07-06
fix(slack-bridge): stop workers taking over Slack threads they don't own (#855)

### LINKED ISSUE #855: slack-bridge: workers can post to Slack threads they don't own (broker-bypass + first-responder-wins races)
## Symptom

Workers repeatedly end up "taking over" Slack threads they were never explicitly assigned to. Recent live observations from a private workspace:

- A worker from a private/internal project posted into a Slack thread it did not own.
- A different worker reported cleanup/done and asked about a thread-ownership handover it had no explicit authorization for.
- **Canonical repro (this issue was itself an instance):** a fresh maintainer DM landed on a worker in a private workspace thread even though the routing task ownership was on a separate Pinet a2a thread. The same message was then redelivered to the worker's Pinet inbox without new signal.

Maintainer report: "this happens all the time … sounds like a routing bug."

## Root cause: several first-responder-wins ownership races + one broker-bypass fallback

This is not one bug. It is the sum of five ownership surfaces that each independently allow "any worker → this thread is now yours":

1. **`sendBrokerMessage` implicitly claims unowned threads.** `@pinet/broker-core/message-send.ts` rejects a send when `thread.ownerAgent && thread.ownerAgent !== senderAgentId`, but if `ownerAgent` is null (or the thread doesn't exist yet in the broker DB) it silently `claimThread(threadId, senderAgentId, …)` and treats first-writer as owner.
2. **`MessageRouter.route` bare-mention claims on released threads.** `@pinet/broker-core/router.ts`, in the "known thread, no owner" branch, falls back to `findAgentMention(text, agents)` after `threadOwnerHint`. Any agent whose name or tail-alias appears in the text can then claim the thread. This is the direct mechanism behind issue #811.
3. **`THREAD_RETARGET_REGEX` is very broad.** Phrases like "pick this up", "take over", "hand this to X", "you take this", "route this to X" combined with a bare agent mention trigger an explicit ownership retarget (`ownerBinding: "explicit"`). Human-to-human phrasing in Slack can accidentally flip ownership.
4. **`runBrokerMaintenancePass` re-assigns released work to any idle worker.** `@pinet/broker-core/maintenance.ts` calls `db.repairThreadOwnership()` which releases claims for stale/purged agents, then `selectBacklogAssignee(backlog, agentLoads, now, busyAssignmentAgeMs)` picks *any* idle worker (or, after `busyAssignmentAgeMs`, the least-loaded working worker) for unrouted backlog. That worker's first send then wins the claim.
5. **`slackProxy("chat.postMessage")` claim race.** `slack-bridge/broker/adapters/slack.ts` `buildSlackApiCapabilityEffects` returns `effects.claimThread` for the thread_ts, and `broker/socket-server.ts applyAdapterCapabilityEffects` calls `router.claimThread(threadId, state.agentId, …)`. The broker sends the Slack message first, then races to claim — first-responder wins with no pre-check.

**Broker-bypass fallback on top of all of this:**

- **`deliverSlackMessage` in `slack-bridge/slack-tools.ts`** has a Pinet-unavailable fallback that goes straight to `slack("chat.postMessage", getBotToken(), body)`. Commit c377253 blocks the "already owned" error from bypassing, but it does not block the case where `pinetDelivery.isAvailable()` returns false (broker link flapped, follower not yet attached). Any follower session that still holds a Slack bot token can direct-post — no ownership check at all.

## Why worker A / worker B / this thread all took over "without rights"

- Prior owner disconnects (crash, ralph loop, tmux churn). Maintenance eventually calls `repairThreadOwnership` and the Slack thread loses its owner.
- The next inbound Slack message either (a) mentions the taking-over worker by name/tail-alias → `findAgentMention` grants the claim (pathway 2), or (b) it is a plain follow-up that hits `selectBacklogAssignee` → the first idle worker is picked and its first outbound send claims the thread (pathway 4), or (c) the worker hits the `pinetDelivery` fallback and direct-posts via the local bot token — no ownership check at all (broker-bypass).
- The Slack API itself has no thread-ownership concept, so as long as the bot token is valid, any session can post anywhere.

## Files involved

- `slack-bridge/node_modules/@pinet/broker-core/router.ts` (routing precedence, `THREAD_RETARGET_REGEX`, `findAgentMention`)
- `slack-bridge/node_modules/@pinet/broker-core/message-send.ts` (implicit claim on unowned thread)
- `slack-bridge/node_modules/@pinet/broker-core/maintenance.ts` (`repairThreadOwnership` + backlog reassignment)
- `slack-bridge/broker/adapters/slack.ts` + `slack-bridge/broker/socket-server.ts` (`slackProxy → chat.postMessage → claimThread` capability effect race)
- `slack-bridge/slack-tools.ts` `deliverSlackMessage` (broker-bypass fallback path)

## Related issues

- #811 (mentions in owned threads over-route)
- #835 (Slack @-mention target identity resolution)
- #772 (Pinet neutrality / Slack coupling review)
- #547 (workspace/install authorization for Slack ingress)

### Phase 1 (this issue): plug the two easiest broker-bypass holes

**A. Block the direct-Slack fallback when Pinet is enabled but the broker is unavailable.**
In `deliverSlackMessage` (`slack-bridge/slack-tools.ts`), if `pinetDelivery` is configured (pinet enabled) AND `pinetDelivery.isAvailable()` returns false, THROW an explicit `broker unavailable, cannot post ownership-guarded Slack message` error instead of silently falling back to `slack("chat.postMessage", getBotToken(), body)`. Only single-player mode (pinetEnabled=false) should ever hit the direct path.

Tests mirror the c377253 shape: given `pinetDelivery.isAvailable() === false` and `threadTs` present, `deliverSlackMessage` throws; the tool surfaces a clear error to the caller instead of posting.

**B. Refuse cross-owner `slackProxy chat.postMessage` in the broker.**
In `broker/adapters/slack.ts` (or upstream in `dispatchAdapterCapability` in `broker/socket-server.ts`), before sending the Slack message, look up the existing thread `ownerAgent`. If it exists and is different from `state.agentId`, refuse with a `thread_not_owned` RPC error so the follower gets a clean error to report and the broker does not race a hostile claim.

Tests: add cases in `broker/adapters/slack.test.ts` / `broker/integration.test.ts` covering (i) unowned thread → post + claim (unchanged behavior), (ii) same-owner thread → post + no-op claim (unchanged), (iii) cross-owner thread → refuse before Slack call, no claim, no adapter effect.

Estimated diff: ~40 LOC + tests per change.

### Phase 2 (follow-up issues to file after Phase 1 lands)

**C. Stop `findAgentMention` from silently claiming a released known thread.** Routing path 3 (thread exists, `ownerAgent === null`, no `threadOwnerHint`) should require either an explicit retarget directive, a `threadOwnerHint` match, or a broker-initiated transfer. Bare mentions in an existing thread should not implicitly claim. Aligns with #811.

**D. Tighten `repairThreadOwnership` semantics.** On stale/purge, mark the thread `orphaned` rather than clearing the owner outright, and refuse implicit re-claim until a human/broker explicitly retargets it. The existing broker `pinet send transfer_thread_id=…` path already covers explicit transfers.

**E. Narrow `THREAD_RETARGET_REGEX`.** Require a stronger signal (leading token, explicit `@bot` mention, or a dedicated slash command) so casual human-to-human phrasing does not flip ownership.

## Acceptance criteria (Phase 1)

- Followers with a valid Slack bot token can NO LONGER post to a Slack thread when Pinet is enabled and the broker is unreachable — they get a clear "broker unavailable" error instead of silently bypassing ownership.
- Followers cannot claim a Slack thread out from under an existing owner by racing a `slackProxy chat.postMessage` — the broker refuses the send before hitting Slack when the thread already has a different owner.
- No behavioural change for the legitimate paths: single-player mode, same-owner sends, unowned first-send claims, and cross-adapter iMessage sends.
- New tests cover both new refusal branches and preserve existing tests.

## Deferred (not in Phase 1)

- Any change to routing precedence (`findAgentMention`, `THREAD_RETARGET_REGEX`, backlog assignment). These land as separate PRs (C/D/E above) once Phase 1 gives us a clean baseline to observe residual takeovers.

gugu91:
Fixes #855

## Problem

Workers repeatedly ended up "taking over" Slack threads they were never explicitly assigned to across recent live observations. This is not a single bug — it's the sum of several first-responder-wins ownership races combined with a broker-bypass fallback (see issue #855 for the full trace).

This PR is Phase 1: plug the two easiest broker-bypass holes without changing routing semantics. Phase 2 issues (findAgentMention silent claim, `repairThreadOwnership`, `THREAD_RETARGET_REGEX`, …) will follow separately.

## Change A — refuse direct-Slack fallback when Pinet is enabled

`slack-bridge/slack-tools.ts` `deliverSlackMessage` previously fell back to `slack("chat.postMessage", getBotToken(), body)` whenever `pinetDelivery.isAvailable()` returned false. Any follower session that still held a Slack bot token could bypass broker ownership entirely — the exact takeover pattern reported.

- The `SlackPinetDeliveryPort` gains an `isEnabled()` signal distinct from `isAvailable()`. `isEnabled()` returns true whenever Pinet is configured for this session (regardless of live broker connectivity).
- `deliverSlackMessage` now refuses the direct-Slack path for threaded replies when `isEnabled() === true`. Single-player mode (`isEnabled() === false`) keeps the direct path exactly as before — no behaviour change there.
- The `isPinetDeliveryFallbackError` heuristic (from c377253) and the `fallbackReason` field are removed: for a threaded reply under Pinet, any Pinet-side error is a hard error — the caller sees the clear message rather than a silent bypass.

New tests in `slack-tools.test.ts`:
- Broker unavailable → threaded `slack_send` throws instead of direct-posting.
- Pinet send error → propagates instead of falling back.
- Pinet timeout → propagates instead of falling back.
- Existing "does not bypass healthy Pinet ownership rejections" test is retained.

Two existing single-player tests (rich blocks, threaded file upload) are updated to use the new `setPinetDeliveryEnabled(false)` toggle so they exercise the sanctioned single-player direct path.

## Change B — refuse cross-owner `slackProxy chat.postMessage` at the broker

`slack-bridge/broker/socket-server.ts` `dispatchAdapterCapability` previously called the Slack adapter first and then raced a first-responder-wins claim via `effects.claimThread`. If a follower slackProxy'd `chat.postMessage` at a thread whose owner had been cleared (stale/purged), the follower took over.

- New `checkAdapterCapabilityThreadOwnership` helper inspects the request pre-adapter. For `slack` + `api.call` + `method: chat.postMessage`, if `params.thread_ts` maps to a broker DB thread with a different `ownerAgent`, the request is rejected before Slack is touched.
- `handleAdapterCapability` and `handleLegacySlackProxy` now require a registered caller (matching every other ownership-sensitive handler); `checkAdapterCapabilityThreadOwnership` keeps a defense-in-depth refusal for threaded `chat.postMessage` from an unregistered caller.
- All other adapters/capabilities and unowned or same-owner threads are unaffected.

New tests in `broker/integration.test.ts`:
- Cross-owner `slackProxy chat.postMessage` throws, does not invoke the adapter, and leaves ownership intact.
- Same-owner `slackProxy chat.postMessage` continues to succeed and preserves the existing owner.
- Unregistered caller `slackProxy` / `invokeAdapterCapability` chat.postMessage refused with `Not registered`; adapter not invoked; pre-owned thread unchanged.

## Prompt guidance

`slack-tools.ts buildSlackSendPromptGuidelines` and `pinet-mesh-ops.ts appendSlackThreadTransferNotice` now tell agents to wait for the broker or ask for a transfer when `slack_send` returns `broker is unavailable` or `already owned by another agent`, instead of retrying via `post_channel` and re-opening the bypass.

## Test/lint/typecheck

- `pnpm lint` ✅
- `pnpm typecheck` ✅
- `pnpm test` — 1462 passing, 1 pre-existing unrelated failure (`pinet-tools.test.ts > formats Pinet dispatcher results for human-readable TUI display`), confirmed by stashing this branch and re-running against main.

## Deferred to follow-up issues

- **C.** Stop `findAgentMention` from silently claiming a released known thread (aligns with #811).
- **D.** Tighten `repairThreadOwnership` semantics — mark orphaned instead of clearing, require explicit retarget.
- **E.** Narrow `THREAD_RETARGET_REGEX` so casual human-to-human phrasing does not flip ownership.

## Repro

Reproduced live during the investigation itself: a private-workspace maintainer DM landed on a worker without an explicit assignment, then the same message was redelivered without new signal. Under this PR, either the broker refuses the send with a clear error, or the message never reaches a worker without a legitimate Pinet-brokered assignment.

## Rollback

If this proves too strict for real ops (e.g. a legitimate follower still needs direct posting during a broker outage), the safest interim is to remove Change A only — the broker-side check in Change B is orthogonal and much less likely to affect normal traffic.

### DIFF
--- a/slack-bridge/broker/integration.test.ts
+++ b/slack-bridge/broker/integration.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, beforeEach, afterEach } from "vitest";
+import { describe, it, expect, beforeEach, afterEach, vi } from "vitest";
 import * as fs from "node:fs";
 import * as os from "node:os";
 import * as path from "node:path";
@@ -1510,6 +1510,128 @@ describe("broker integration — client ↔ server ↔ DB", () => {
 
     // No thread should be created
     expect(db.getThread("some-ts")).toBeNull();
+  });
+
+  it("refuses slackProxy chat.postMessage to a thread owned by a different agent (#855)", async () => {
+    client.disconnect();
+    await server.stop();
+
+    const invokeCapability = vi.fn(async ({ params }: { params: Record<string, unknown> }) => {
+      const body = params.params as Record<string, unknown>;
+      return {
+        result: {
+          ok: true,
+          ts: "unauthorized-reply-ts",
+          channel: body.channel,
+          message: { ts: "unauthorized-reply-ts", text: body.text },
+        },
+        effects: {
+          claimThread: {
+            threadId: "protected-thread-ts",
+            channel: body.channel as string,
+          },
+        },
+      };
+    });
+
+    server = new BrokerSocketServer(db, { type: "tcp", host: "127.0.0.1", port: 0 });
+    server.setOutboundMessageAdapters([
+      {
+        name: "slack",
+        send: async () => undefined,
+        invokeCapability,
+      },
+    ]);
+    await server.start();
+
+    const info = server.getConnectInfo();
+    if (info.type !== "tcp") throw new Error("Expected TCP");
+
+    // Pre-seed the thread as owned by a different agent.
+    db.registerAgent("legit-owner", "Legit Owner", "🛡️", process.pid);
+    db.createThread({
+      threadId: "protected-thread-ts",
+      source: "slack",
+      channel: "C-PROTECTED",
+      ownerAgent: "legit-owner",
+      createdAt: new Date().toISOString(),
+      updatedAt: new Date().toISOString(),
+    });
+
+    client = new BrokerClient({ host: info.host, port: info.port });
+    await client.connect();
+    await client.register("would-be-hijacker", "🥷");
+
+    await expect(
+      client.slackProxy("chat.postMessage", {
+        channel: "C-PROTECTED",
+        text: "trying to hijack",
+        thread_ts: "protected-thread-ts",
+      }),
+    ).rejects.toThrow(/already owned by another agent/i);
+
+    // Adapter must NOT have been called: refusal happens before hitting Slack.
+    expect(invokeCapability).not.toHaveBeenCalled();
+
+    // Ownership must be unchanged after the refused call.
+    const thread = db.getThread("protected-thread-ts");
+    expect(thread?.ownerAgent).toBe("legit-owner");
+  });
+
+  it("allows slackProxy chat.postMessage when caller is the current thread owner (#855)", async () => {
+    client.disconnect();
+    await server.stop();
+
+    const invokeCapability = vi.fn(async ({ params }: { params: Record<string, unknown> }) => {
+      const body = params.params as Record<string, unknown>;
+      return {
+        result: {
+          ok: true,
+          ts: "owner-reply-ts",
+          channel: body.channel,
+          message: { ts: "owner-reply-ts", text: body.text },
+        },
+        effects: {
+          claimThread: { threadId: "owned-thread-ts", channel: body.channel as string },
+        },
+      };
+    });
+
+    server = new BrokerSocketServer(db, { type: "tcp", host: "127.0.0.1", port: 0 });
+    server.setOutboundMessageAdapters([
+      {
+        name: "slack",
+        send: async () => undefined,
+        invokeCapability,
+      },
+    ]);
+    await server.start();
+
+    const info = server.getConnectInfo();
+    if (info.type !== "tcp") throw new Error("Expected TCP");
+
+    client = new BrokerClient({ host: info.host, port: info.port });
+    await client.connect();
+    const reg = await client.register("real-owner", "✅");
+
+    db.createThread({
+      threadId: "owned-thread-ts",
+      source: "slack",
+      channel: "C-OWNED",
+      ownerAgent: reg.agentId,
+      createdAt: new Date().toISOString(),
+      updatedAt: new Date().toISOString(),
+    });
+
+    await client.slackProxy("chat.postMessage", {
+      channel: "C-OWNED",
+      text: "legit reply",
+      thread_ts: "owned-thread-ts",
+    });
+
+    expect(invokeCapability).toHaveBeenCalledOnce();
+    const thread = db.getThread("owned-thread-ts");
+    expect(thread?.ownerAgent).toBe(reg.agentId);
   });
 });
 
--- a/slack-bridge/broker/integration.test.ts
+++ b/slack-bridge/broker/integration.test.ts
@@ -1633,6 +1633,61 @@ describe("broker integration — client ↔ server ↔ DB", () => {
     const thread = db.getThread("owned-thread-ts");
     expect(thread?.ownerAgent).toBe(reg.agentId);
   });
+
+  it("refuses adapter.capability from an unregistered caller (#855)", async () => {
+    client.disconnect();
+    await server.stop();
+
+    const invokeCapability = vi.fn(async () => ({
+      result: { ok: true },
+    }));
+
+    server = new BrokerSocketServer(db, { type: "tcp", host: "127.0.0.1", port: 0 });
+    server.setOutboundMessageAdapters([
+      {
+        name: "slack",
+        send: async () => undefined,
+        invokeCapability,
+      },
+    ]);
+    await server.start();
+
+    const info = server.getConnectInfo();
+    if (info.type !== "tcp") throw new Error("Expected TCP");
+
+    // Pre-seed an owned thread that the unregistered caller must not touch.
+    db.registerAgent("legit-owner-2", "Legit Owner 2", "🛡️", process.pid);
+    db.createThread({
+      threadId: "pre-existing-thread",
+      source: "slack",
+      channel: "C-EXISTING",
+      ownerAgent: "legit-owner-2",
+      createdAt: new Date().toISOString(),
+      updatedAt: new Date().toISOString(),
+    });
+
+    client = new BrokerClient({ host: info.host, port: info.port });
+    await client.connect();
+    // Deliberately DO NOT call client.register(...).
+
+    await expect(
+      client.slackProxy("chat.postMessage", {
+        channel: "C-EXISTING",
+        text: "unregistered hijack attempt",
+        thread_ts: "pre-existing-thread",
+      }),
+    ).rejects.toThrow(/not registered/i);
+
+    await expect(
+      client.invokeAdapterCapability("slack", "api.call", {
+        method: "chat.postMessage",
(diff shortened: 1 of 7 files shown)


will-jack-and-jilll (commented):
*Will's reviewer agent Race Stoat*

## Verdict
- **Block**
- Change A looks good, but Change B still leaves an unregistered-caller escape hatch.

## Top finding
- `slack-bridge/broker/socket-server.ts:1396-1522` — `adapter.capability` / legacy `slack.proxy` still do not require registration, and `checkAdapterCapabilityThreadOwnership()` immediately returns when `state.agentId` is null. A connected-but-unregistered caller can therefore still send `slack api.call chat.postMessage` into another agent’s thread, and the message reaches Slack before any ownership enforcement.
  - Suggested fix: require registration for `adapter.capability` / `slack.proxy`, or at minimum reject threaded `chat.postMessage` when `state.agentId` is missing. Please add an integration test for the unregistered caller case.

## Verification
- `cd .worktrees/fix-thread-ownership-855/slack-bridge && pnpm exec vitest run slack-tools.test.ts broker/integration.test.ts` ✅ (143 tests)
- `cd slack-bridge && pnpm exec vitest run slack-tools.test.ts broker/integration.test.ts` ✅ (140 tests on main baseline)
- `cd .worktrees/fix-thread-ownership-855/slack-bridge && pnpm exec vitest run pinet-tools.test.ts -t "formats Pinet dispatcher results for human-readable TUI display"` ❌ pre-existing unrelated failure; this PR does not touch `pinet-tools.ts` / `pinet-tools.test.ts`

gugu91:
@code-reviewer follow-up: addressed Race Stoat blocker in 5144bbe.

- `handleAdapterCapability` and `handleLegacySlackProxy` now require a registered caller (`Not registered` error), matching every other ownership-sensitive handler.
- `checkAdapterCapabilityThreadOwnership` keeps a defense-in-depth refusal for threaded `chat.postMessage` from an unregistered caller, in case a future call path forgets the handler-level guard.
- New integration test `refuses adapter.capability from an unregistered caller (#855)` exercises `client.connect()` **without** `register()` and asserts both `slackProxy` and `invokeAdapterCapability` threaded `chat.postMessage` are refused, the adapter is not invoked, and the pre-owned thread is unchanged.
- Prompt guidance in `slack-tools.ts buildSlackSendPromptGuidelines` and `pinet-mesh-ops.ts appendSlackThreadTransferNotice` now tells agents to wait for the broker (or ask for a transfer) on `broker is unavailable` / `already owned by another agent` instead of falling back to `post_channel`.

Pre-existing `pinet-tools.test.ts > formats Pinet dispatcher results for human-readable TUI display` failure confirmed present on main as well.

PR #864, opened 2026-07-07
feat(transport-core): extract shared async primitives (sleep/withTimeout/backoff)

tmustier:
## Why

Follow-up to a "should we adopt Effect?" review: the answer was no (conflicts with the zero-runtime-dep convention, viral abstraction, interop friction at the pi ExtensionAPI boundary), but the review surfaced real duplication in hand-rolled async primitives across packages. This PR extracts those primitives into a shared zero-dependency module instead.

Survey findings: there are **no generic retry loops** in the repo — each retry driver (Slack 429, broker reconnect, socket-mode reconnect) is legitimately bespoke. The duplication is in the *primitives* they compose. So this ships exactly those, and deliberately **skips** a speculative `retry()`/`Semaphore` (zero call sites today).

## What

New `@pinet/transport-core/async` subpath export (mirrors the pinet-core subpath pattern):

- `sleep(ms, { signal? })` — abortable delay, rejects with `AbortError`
- `createAbortError` / `isAbortError` — moved from slack-bridge helpers
- `withTimeout(promise, ms, label?)` — unref'd guard timer, `TimeoutError`, normalizes non-Error rejections (behavior parity with the broker/client copy, plus an error name)
- `computeBackoffDelay(attempt, { initialMs, maxMs, factor?, jitterRatio?, random? })` — capped jittered exponential backoff; exact formula parity with `computeReconnectDelay` (test included)

Migrated call sites (packages already depending on transport-core):

- `slack-bridge/helpers.ts` — `abortableDelay` delegates to `sleep`; abort helpers re-exported (no churn for existing importers); `callSlackAPI` 429 wait unified for signal/no-signal branches
- `slack-bridge/broker/client.ts` — local `withTimeout` removed; `computeReconnectDelay` delegates to `computeBackoffDelay` (existing tests pass unchanged)
- `slack-bridge/subtree-broker-runtime.ts` — local `delay()` replaced

Deliberately **not** migrated: `neon-psql`/`browser-playwright` inline delays (would add a workspace dep for one `sleep(100)`).

## Notes for review

- Rebased onto `main` post-0.2.4/#861. Passes the new `agent-standards-lint`: tests use node:assert object/function predicates instead of `(error: unknown)` callbacks, and net explicit-`unknown` count across changed files is flat (the two remaining additions are the `isAbortError`/`isTimeoutError` type-guard signatures, one of which is the pre-existing guard moved from helpers).
- An earlier commit here fixed the stale `2026-07-01` schedule-test timestamp; dropped after `8c3d21a` landed the freeze-clock fix on `main`.
- 21 node:test cases in `transport-core/async.test.ts` (transport-core convention), including a parity test against the historical reconnect-delay formula.
- Root `tsconfig.json` gains `@pinet/transport-core/*` path mapping for source-level typecheck; runtime resolves via the new `./async` export in `transport-core/package.json`.
- Zero token-footprint impact: no new tools, prompts, or schemas.
- Full gate green: `pnpm lint` (incl. agent-standards), `pnpm typecheck`, `pnpm test` (11/11 tasks), `pnpm format:check`.

### DIFF
--- a/slack-bridge/broker/client.ts
+++ b/slack-bridge/broker/client.ts
@@ -1,4 +1,5 @@
 import * as net from "node:net";
+import { computeBackoffDelay, withTimeout } from "@pinet/transport-core/async";
 import { readMeshSecret } from "./auth.js";
 import { DEFAULT_SOCKET_PATH as PINET_DEFAULT_SOCKET_PATH } from "./paths.js";
 import { assertLoopbackTcpHost } from "./raw-tcp-loopback.js";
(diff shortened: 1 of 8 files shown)


tmustier:
**pr-reviewer (gpt-5.5 xhigh) findings:**

Code looks good. No P0–P2 issues found.

**Verdict: correct** — the async primitive extraction and migrated call sites appear sound.

Local gate: `pnpm lint` / `pnpm typecheck` / `pnpm test` all 11/11 tasks green, `pnpm format:check` clean.

tmustier:
**pr-reviewer (gpt-5.5 xhigh), post-rebase re-review:** No P0–P2 findings. **Verdict: correct** — shared async primitives preserve existing slack-bridge behavior; subpath export and test coverage sound. CI quality check green.

PR #935, opened 2026-07-13
fix(slack-bridge): preserve prompt cache across Pinet role changes

### LINKED ISSUE #934: Keep Pinet runtime guidance prompt-cache stable across role changes
## Problem

`/pinet follow` changes the extension-owned suffix of Pi's system prompt by adding follower workflow guidance and registration-derived identity/personality guidance. Because provider prompt caches require an exact stable prefix, the role transition can invalidate the cached system block even though Pi's base system prompt is unchanged.

Broker transitions have the same underlying risk because broker prompt MD and mutable identity/skin guidance currently enter the system prompt.

## Proposed behavior

- Keep the system prompt byte-stable across Pinet role, identity, skin, and broker-prompt changes.
- Keep role-invariant safety boundaries in the system prompt.
- Deliver mutable runtime guidance through hidden durable extension context using Pi's public `sendMessage` API.
- Append initial guidance before the first agent turn and updated snapshots after role/identity changes.
- Reinject current guidance after compaction.
- Serialize updates, prevent old-session queued work from leaking into replacement sessions, and deduplicate only after successful delivery.

## Why this shape

Pi does not expose a portable extension hook for adding mid-conversation system/developer messages. Custom extension messages are durably stored and converted to ordinary user-role context by Pi's provider adapters. Security-critical broker restrictions therefore remain stable system instructions and runtime-enforced tool restrictions; mutable operational guidance can safely use appended context.

## Acceptance criteria

- [ ] Full system prompt is byte-identical before and after `/pinet follow`.
- [ ] Registration identity changes do not alter the system prompt.
- [ ] Follower and broker workflow guidance is available before the next relevant turn.
- [ ] Role deactivation appends an authoritative inactive snapshot.
- [ ] Compaction restores current runtime guidance.
- [ ] Concurrent updates are ordered and session replacement invalidates queued old-session updates.
- [ ] Slack bridge tests cover prompt stability and role-context behavior.

tmustier:
## Summary

- keep the extension-owned system-prompt suffix byte-stable across Pinet role and identity changes
- move mutable identity, skin, broker MD, and follower workflow guidance into hidden durable Pi context
- retain role-invariant broker safety boundaries at system priority and in runtime tool enforcement
- serialize/deduplicate runtime guidance updates, fence queued work across session replacement, restore guidance after compaction, and await follower identity updates

Closes #934.

## Design note

Pi has no portable public extension hook for appending mid-conversation system/developer messages. `pi.sendMessage()` creates durable custom messages that Pi converts to ordinary user-role provider context. Mutable operational guidance therefore uses that path, while security-critical broker restrictions remain in the stable system prompt and runtime guardrails.

## Validation

- `pnpm prepush`
- Slack bridge: 89 test files passed, 2 skipped; 1,571 tests passed, 3 skipped
- prompt stability integration test confirms the complete system prompt is byte-identical before and after broker activation
- follower tests confirm identity is applied and guidance appended before registration/reconnect processing continues

### DIFF
--- a/slack-bridge/agent-prompt-guidance.test.ts
+++ b/slack-bridge/agent-prompt-guidance.test.ts
@@ -26,33 +26,35 @@ function createDeps(overrides: Partial<AgentPromptGuidanceDeps> = {}): AgentProm
 }
 
 describe("createAgentPromptGuidance", () => {
-  it("appends identity, personality, and reaction guidance for non-mesh sessions", async () => {
+  it("keeps only role-invariant reaction and broker guardrails in the system prompt", async () => {
     const getIdentityGuidelines = vi.fn(() => ["IDENTITY 1", "IDENTITY 2", "IDENTITY 3"]);
-    const guidance = createAgentPromptGuidance(
-      createDeps({
-        getIdentityGuidelines,
-      }),
-    );
+    const guidance = createAgentPromptGuidance(createDeps({ getIdentityGuidelines }));
 
     const result = await guidance.beforeAgentStart({ systemPrompt: "BASE" });
 
-    expect(getIdentityGuidelines).toHaveBeenCalledTimes(1);
-    expect(result.systemPrompt).toContain("BASE\n\nIDENTITY 1\nIDENTITY 2\nIDENTITY 3");
-    expect(result.systemPrompt).toContain("COMMUNICATION STYLE:");
-    expect(result.systemPrompt).toContain("For `Cobalt Olive Crane`, aim for a");
-    expect(result.systemPrompt).toContain("Slack emoji reactions are ignored by default");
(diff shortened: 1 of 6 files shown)


tmustier:
Withdrawing this approach: the implementation is not acceptable as written.

It moves mutable operational control text into hidden, durable user-role messages. That pollutes model context, consumes tokens before the first user message, leaves superseded guidance in session history, and weakens the instruction priority of operational guidance.

The follow-up commit that suppresses the initial off-state message addresses one symptom, not the underlying design. Pinet being off must mean no Pinet context. When roles change, correct prompt semantics and transparent context should take precedence over preserving the provider prompt cache.

Please do not merge this PR. This needs a redesign rather than hidden control messages.

PR #952, opened 2026-07-25
feat: stranded-broker recovery — lock owner identity, conflict classification, and /pinet start replace

### LINKED ISSUE #951: Pinet broker ownership can become stranded after its controlling Pi session is lost
## Summary

If the Pi session that started the Pinet broker is lost, the broker can become operationally stranded: a process still holds the singleton lock, database, and Unix socket, but the operator no longer has the broker's Pi session and a new session cannot take over.

Running `/pinet start` in the new Pi session fails with:

```text
Pinet broker failed: Another pinet broker is already running. Only one broker may be active at a time.
```

The single-broker guard is correct and should remain fail-safe. The problem is that this error collapses several materially different states into one and provides no recovery path:

1. a healthy broker that the new session should follow;
2. a dead broker with stale lock/socket artifacts;
3. a live but stalled or operationally orphaned broker process;
4. a stale PID lock whose PID has been reused by an unrelated process.

## User context

I hit this repeatedly after losing the active broker's Pi session, including around laptop restarts, crashes, and stalled processes. From the operator's point of view there is no usable broker session, but `/pinet start` says a broker is already running.

This turns routine recovery into manual process forensics and makes it easy to either leave Pinet unavailable or kill the wrong process while trying to recover.

## Captured incident

In the latest incident:

- A new Pi session was not connected to Pinet. Pinet commands reported that Pinet was not running locally.
- `/pinet start` failed with the singleton-broker error above.
- `~/.pi/pinet-broker.lock` contained PID `1336`.
- PID `1336` was alive.
- `lsof` showed that PID `1336` had `~/.pi/pinet-broker.db`, its WAL/SHM files, and `~/.pi/pinet.sock` open.
- The original interactive broker session was no longer available to the operator.
- Sending `SIGTERM` to PID `1336` made it exit cleanly and removed both the lock and socket.
- After that, a fresh broker could be started normally.

So this was not the simple stale-lock case where the recorded PID was dead. A real process still owned the broker resources, but broker ownership/control was stranded from the user's perspective.

## Current implementation context

The leader lock currently stores only a PID and treats `kill(pid, 0)` as sufficient proof that another broker is live:

https://github.com/gugu91/pinet/blob/bf434ebdcdc00511d05dffc01994ada15b4f13ba/broker-core/leader.ts#L9-L44

Broker startup then turns any lock-acquisition failure into the same generic error:

https://github.com/gugu91/pinet/blob/bf434ebdcdc00511d05dffc01994ada15b4f13ba/slack-bridge/broker/index.ts#L65-L102

That is enough to prevent split-brain, but not enough to determine whether the lock owner is healthy, reachable, still attached to an operator session, or even the same process that originally wrote the lock.

## Expected operator experience

When `/pinet start` encounters an existing lock, it should preserve the single-broker invariant while diagnosing the owner and presenting a safe next step.

At minimum, distinguish:

- **Healthy, reachable broker:** report that a broker is already active, identify it safely, and suggest or offer `/pinet follow` rather than presenting a generic failure.
- **Dead or invalid lock owner:** reclaim stale state and start normally.
- **Live PID but failed broker health probe:** report a stalled/stranded broker and provide an explicit, fenced recovery path.

`/pinet status` should also be able to inspect machine-wide broker state from a session whose local Pinet runtime is off. Today that is precisely the session from which recovery is needed.

## Suggested design direction, not a prescribed implementation

A robust solution likely needs more than PID liveness alone:

- Store structured lock identity, for example PID, process start identity, broker instance ID, Pi session ID, and start time.
- Probe the broker socket with a bounded health/status RPC before classifying it as healthy.
- Verify that the lock owner matches the process/socket expected for that broker instance, avoiding PID-reuse false positives.
- Add an explicit operator action such as `/pinet restart` or `/pinet start --replace` for the unhealthy/stranded case.
- Make takeover conservative: try graceful RPC shutdown first, then bounded termination only after re-verifying the same fenced owner. Never silently kill an arbitrary live PID.
- If the existing broker is healthy, prefer follow/attach guidance rather than replacement.

The key behavior is that ordinary recovery should not require reading a lock file, running `lsof`, and manually signaling a PID.

## Related issues

- #119 / PR #145 introduced the singleton PID lock to prevent split-brain. This issue should preserve that safety property.
- #368 described a dead PID plus stale lock/socket, but it was closed after that incident was found to be misdiagnosed. The incident here is different: the recorded PID was alive and owned the broker resources.
- #936 covers ordinary sessions inheriting global broker mode and noisily attempting singleton startup. This issue concerns explicit recovery after the controlling broker session is lost.
- #420 documented the broader daemon/control-plane lifecycle direction. This issue does not require daemonization; it asks for safe discovery and recovery in the current embedded-broker architecture.

## Acceptance sketch

- A second session can distinguish a healthy existing broker from a stale, invalid, or unhealthy owner.
- Healthy-broker collisions produce actionable follow/attach guidance, not a generic failure.
- A stranded/unhealthy broker has a safe explicit restart/takeover path that preserves split-brain protection.
- `/pinet status` can report global broker ownership and health even when the current session is not connected.
- Lock identity is not validated by PID existence alone.
- Regression coverage includes dead PID, PID reuse/unrelated live PID, healthy broker, unresponsive live broker, graceful replacement, and failed replacement without split-brain.

tmustier:
Closes #951.

## Problem

When the Pi session hosting the broker is lost (crash, laptop restart, stalled process), `/pinet start` collapses every leader-lock conflict into one generic error with no recovery path. The lock stores only a PID validated by `kill(pid, 0)`, so a second session cannot distinguish a healthy broker to follow, a dead broker with stale state, a stranded live process, or a reused PID. Recovery required manual lock forensics (`cat` the lock, `lsof`, signal a PID by hand).

## What this PR does

Design notes: `plans/951-stranded-broker-recovery.md`.

### 1. Structured leader lock (`broker-core/leader.ts`)

- Lock format: PID on line 1 (legacy readers still `parseInt` it, preserving mixed-version single-broker safety) + a JSON metadata line with process start time, per-acquisition instance id, hostname, and creation timestamp.
- New inspection API (`readBrokerLockOwner`, `inspectBrokerLock`, `getProcessStartTime`) classifies the owner: `none | unreadable | stale-dead | stale-pid-reused | alive`.
- PID reuse is only declared when both recorded and current start times are known and differ — uncertainty never reclaims a live broker's lock. Start times are captured deterministically (`/proc/<pid>/stat` on Linux, `LC_ALL=C ps -o lstart=` elsewhere) and compared by exact equality.

### 2. Conflict classification (`slack-bridge/broker/lock-conflict.ts`)

- `startBroker` classifies lock conflicts: lock inspection + a bounded socket probe (any well-formed JSON-RPC response, including an auth error, proves the broker event loop is serving).
- Throws a typed `BrokerLockConflictError`: an active broker suggests `/pinet follow` or `/pinet start replace`; a stranded owner suggests `/pinet start replace`. Reclaimable conflicts (owner died mid-start) retry the acquire once.

### 3. Graceful remote shutdown (`admin.shutdown` RPC)

- New authenticated socket-server RPC. The broker runtime wires it to stop the Pinet runtime in the owning session — the session survives, it just stops being the broker, with an operator notification.
- Brokers without a wired handler report method-not-found, so callers treat them as `unsupported` and fall back.

### 4. Conservative takeover (`/pinet start replace`)

`replaceBrokerOwner()` sequence:

1. Stale/absent lock → nothing to replace (normal start reclaims).
2. Live owner → request `admin.shutdown`, poll for lock release (bounded).
3. Fallback → re-verify the owner fence (pid + process start time + instance id), SIGTERM the verified owner, poll again (bounded).
4. Never SIGKILL. Aborts if the owner identity changes mid-flight. Refuses to signal the calling process itself.

### 5. Operator surfaces

- `/pinet status` in `off`/`single` modes appends a machine-wide "Global broker" line (lock owner, socket health, concrete next step) — usable from exactly the disconnected session that needs recovery.
- README documents the recovery flow.

## Safety properties preserved

- The single-broker invariant is untouched: `alive` owners are never reclaimed automatically; takeover is an explicit operator action.
- Legacy locks and legacy readers interoperate both ways (covered by tests).
- All fallback termination is fenced on full owner identity and re-verified immediately before signaling.

## Testing

- `pnpm prepush` (lint, typecheck, full test suite) green; 60+ new assertions across `leader.test.ts`, `lock-conflict.test.ts`, `helpers.test.ts`, `pinet-commands.test.ts` covering: v2/legacy lock parsing, dead-PID and PID-reuse reclaim, unknown-start-time conservatism, probe healthy/unreachable/unresponsive (against a real `BrokerSocketServer`), `admin.shutdown` auth gating and unsupported fallback, replace graceful/terminated/owner-changed/failed/self-PID paths, and the new command/status surfaces.
- Cross-process end-to-end runs (real broker child process on temp paths):
  - graceful: `probe healthy → admin.shutdown accepted → lock released → fresh broker starts`
  - stubborn (no handler): `unsupported → fenced SIGTERM → lock released → fresh broker starts`
  - The E2E caught and fixed a real bug: the replace poll loop's default sleep timer was unref'd, letting a bare event loop exit mid-poll.

### DIFF
--- a/broker-core/leader.ts
+++ b/broker-core/leader.ts
@@ -58,6 +58,12 @@ export interface BrokerLockProbes {
  * for exact equality against a value captured by this same function, so the
  * format does not need to be parseable — only stable for a given process.
  *
+ * On Linux the tick count is only unique within a single boot, so it is
+ * scoped with the kernel boot id: a PID reused after a reboot can then never
+ * present the same identity as the pre-reboot owner. `ps lstart` has
+ * one-second resolution, so a PID reused within the same wall-clock second
+ * as the original process is indistinguishable — an accepted residual risk.
+ *
  * Returns null when the start time cannot be determined; callers must treat
  * null as "unknown" and never use it as evidence of staleness.
  */
@@ -77,7 +83,17 @@ export function getProcessStartTime(pid: number): string | null {
         // Fields after comm+state start at index 1 here; starttime is overall
         // field 22, i.e. index 19 of the post-state remainder.
         const startTime = fields[19];
-        if (startTime && /^\d+$/.test(startTime)) return startTime;
+        if (startTime && /^\d+$/.test(startTime)) {
+          // starttime is measured in clock ticks since boot — scope it with
+          // the boot id so identities never collide across reboots.
+          let bootId = "";
+          try {
+            bootId = fs.readFileSync("/proc/sys/kernel/random/boot_id", "utf-8").trim();
+          } catch {
+            /* boot id unavailable — token stays valid within this boot */
+          }
+          return bootId ? `${bootId}:${startTime}` : startTime;
+        }
       }
     } catch {
       /* fall through to ps */
@@ -259,6 +275,15 @@ function exclusiveCreateFile(filePath: string, content: string): boolean {
  * with staleness re-verified under that mutex — so a fresh lock can never be
  * destroyed by a concurrent reclaimer, and the lock path never goes empty
  * while a live owner's lock exists.
+ *
+ * Known mixed-version limitation: builds that predate the structured format
+ * replace a lock they consider stale with a plain rename over the lock path,
+ * which can overwrite a just-acquired v2 lock when an old and a new build
+ * race over the same stale lock. Exclusive acquisition is therefore only
+ * guaranteed among processes running this code; the window disappears once
+ * no pre-v2 sessions remain. A representation old builds cannot overwrite
+ * (such as a lock directory) would also break their ability to read the
+ * owner PID, which this format deliberately preserves.
  */
 export class LeaderLock {
   private readonly lockPath: string;
@@ -341,23 +366,25 @@ export class LeaderLock {
   /**
    * Remove a stale lock under an exclusive reclaim mutex.
    *
-   * The mutex file (`<lockPath>.reclaim`) is created with `O_EXCL`, so at
-   * most one reclaimer proceeds at a time, and staleness is re-verified
-   * while holding it. Because stale locks are only ever unlinked under this
-   * mutex, the lock path cannot go empty while a live owner's lock exists —
-   * which is what makes the exclusive create in `tryAcquire` a sound
-   * arbiter. A mutex left behind by a crashed reclaimer (dead PID) is itself
-   * reclaimed.
+   * The mutex file (`<lockPath>.reclaim`) is created with `O_EXCL` and
+   * records the reclaimer's PID plus start identity, so at most one
+   * reclaimer proceeds at a time, and staleness is re-verified while holding
+   * it. Because stale locks are only ever unlinked under this mutex, the
+   * lock path cannot go empty while a live owner's lock exists — which is
+   * what makes the exclusive create in `tryAcquire` a sound arbiter. A mutex
+   * left behind by a crashed reclaimer (dead PID, or a PID provably reused
+   * by an unrelated process) is itself reclaimed.
    *
    * Returns true when the caller may retry an exclusive create.
    */
   private reclaimStaleLock(): boolean {
     const mutexPath = `${this.lockPath}.reclaim`;
     const isRunning = this.probes.isProcessRunning ?? isProcessRunning;
+    const startTimeOf = this.probes.getProcessStartTime ?? getProcessStartTime;
 
     let holdsMutex = false;
     for (let campaign = 0; campaign < 2 && !holdsMutex; campaign++) {
-      if (exclusiveCreateFile(mutexPath, `${process.pid}\n`)) {
+      if (exclusiveCreateFile(mutexPath, `${process.pid}\n${startTimeOf(process.pid) ?? ""}\n`)) {
         holdsMutex = true;
       } else {
         // Mutex exists — held by a live reclaimer, or left by a crashed one.
@@ -367,9 +394,18 @@ export class LeaderLock {
         } catch {
           continue; // vanished — retry the campaign
         }
-        const holderPid = /^\d+$/.test(holderRaw) ? parseInt(holderRaw, 10) : null;
+        const [pidLine = "", startLine = ""] = holderRaw.split("\n");
+        const holderPid = /^\d+$/.test(pidLine.trim()) ? parseInt(pidLine.trim(), 10) : null;
+        const recordedStart = startLine.trim();
         if (holderPid !== null && holderPid > 0 && isRunning(holderPid)) {
-          return false; // a live reclaimer is working — back off
+          // The PID is live, but it may be an unrelated process that reused
+          // a crashed reclaimer's PID. Back off unless the recorded start
+          // identity provably belongs to a different process — uncertainty
+          // always means "assume the reclaimer is alive".
+          const currentStart = startTimeOf(holderPid);
+          if (!recordedStart || currentStart === null || currentStart === recordedStart) {
+            return false; // a live reclaimer is working — back off
+          }
         }
         // Crashed reclaimer. Remove its mutex only if it still records the
         // dead holder we just read, then retry the campaign.
--- a/broker-core/leader.ts
+++ b/broker-core/leader.ts
@@ -1,3 +1,5 @@
+import { execFileSync } from "node:child_process";
+import * as crypto from "node:crypto";
 import * as fs from "node:fs";
 import * as path from "node:path";
 import * as os from "node:os";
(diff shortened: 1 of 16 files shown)


tmustier:
## Review findings: changes requested

CI is green, but I found two P1 safety issues and one P2 fencing issue that should be fixed before merge.

### 1. P1: simultaneous acquisition can produce multiple leaders

`broker-core/leader.ts:250-288`

Acquisition checks the lock, runs the process-start probe, then overwrites the lock path with `renameSync`. Multiple contenders can all inspect an empty path, each overwrite it, verify before the next overwrite, and return `true`.

I reproduced this with 12 synchronized child processes. In 4 of 5 runs against this PR, 2-3 processes reported successful acquisition. The same harness against the base revision produced exactly one winner in all 5 runs. The new `ps` call between inspection and rename materially expands the race window.

Please use an exclusive OS-level acquisition primitive rather than check-then-rename, including fenced stale-lock reclamation. The post-write verification is not sufficient because each contender can verify before the next overwrite.

### 2. P1: legacy locks can SIGTERM an unrelated reused PID

`slack-bridge/broker/lock-conflict.ts:454-515`

Legacy owners have both `processStartTime` and `instanceId` set to `null`, so `sameOwner()` effectively compares only the PID. If the original broker exits and that PID is reused, `/pinet start replace` can signal the unrelated replacement process.

I reproduced the start identity changing from the old owner to an unrelated process while the legacy lock remained unchanged; `replaceBrokerOwner()` still called `kill(pid, "SIGTERM")` and returned `replaced-terminated`.

Legacy or otherwise unfenced owners should never receive automatic SIGTERM. After graceful RPC failure, abort with manual recovery guidance instead.

### 3. P2: release does not verify the acquisition instance

`broker-core/leader.ts:294-302`

Structured locks introduce a per-acquisition `instanceId`, but `release()` still unlinks any lock whose PID equals `process.pid`. A stale `LeaderLock` object can therefore delete a successor lock acquired by another broker instance in the same process, reopening the split-brain path.

Require the on-disk `instanceId` to equal `this.instanceId` (and preferably verify the recorded process start identity) before unlinking.

tmustier:
All three findings addressed in 32d29d8. Point-by-point:

### 1. Simultaneous acquisition — fixed, with one correction to the analysis

Confirmed and fixed, but the baseline claim doesn't hold: with a fair harness (winners hold the lock until every contender has decided), the **base algorithm admits 4–7 leaders out of 8 contenders** — the race is pre-existing, not introduced by this PR. A harness whose winners exit immediately measures legitimate dead-owner reclamation, not mutual exclusion, which likely explains the "base was always single-winner" observation. Either way this PR is the right place to fix it, since it builds takeover machinery on top of the lock.

The fix in `broker-core/leader.ts`:

- The lock is now created via write-temp + `link()` — atomic, exclusive, and fully written at the instant the path exists — so the kernel picks exactly one winner among simultaneous contenders. (Plain `writeFileSync("wx")` was not enough: it opens then writes, so a concurrent reader can observe an empty just-created lock, classify it unreadable, and reclaim it.)
- Stale locks are only ever unlinked while holding an exclusively-created `<lockPath>.reclaim` mutex, with staleness re-verified under the mutex. This means the lock path can never go empty while a live owner's lock exists, which is what makes the exclusive create a sound arbiter. A mutex left behind by a crashed reclaimer (dead PID) is itself reclaimed.
- A quarantine-rename approach was tried first and rejected: renaming the lock aside exposes an empty path that a third contender can win while the displaced fresh lock sits in quarantine (caught by the new stress test).

Two multi-process regression tests now race 8 real contender processes over an empty path and over a stale lock, asserting exactly one winner — 10/10 consecutive runs green locally.

Known residual: old-build brokers don't speak the reclaim-mutex protocol, so mixed-version simultaneous starts retain the old race until upgraded. Same-version guarantees are unconditional.

### 2. Legacy-lock SIGTERM — fixed

`replaceBrokerOwner` now refuses automatic termination for any owner without a recorded process start identity (legacy locks, or structured locks where the start-time probe failed at write time). The graceful `admin.shutdown` path is still attempted; after it fails, the outcome is `failed` with explicit manual guidance (`ps -p <pid>`, terminate manually, then `/pinet start`). Note the two cases compound safely: old brokers that report `unsupported` for the RPC are the same brokers that wrote legacy locks, so they always land on the manual path. README documents this.

Related hardening from a parallel review pass: `requestBrokerShutdown` now distinguishes `rejected` (well-formed error response — auth rejection, typically a mesh secret mismatch) from `failed` (connected but hung). A responsive broker that rejects shutdown is not stranded, so replacement aborts without signalling instead of falling through to SIGTERM; `failed`/`unreachable` (genuinely hung/dead socket) still proceed to the fenced fallback.

### 3. Release instance fence — fixed

`release()` now requires the on-disk `instanceId` to equal this acquisition's instance id, not just the PID, before unlinking. New regression test covers the same-process successor scenario.

Full `pnpm prepush` green; cross-process E2E re-verified (graceful `admin.shutdown` replacement and fenced SIGTERM replacement both succeed with the new lock).

tmustier:
## Second review round (GPT-5.6 adversarial pass) — triage and resolution

An independent adversarial review of `32d29d8` produced six findings. Triage:

| Finding | Severity | Resolution |
|---|---|---|
| Linux `/proc` starttime is boot-relative — a PID reused after reboot could present the pre-reboot owner's identity to the SIGTERM fence | P1 | **Fixed** in `ed595dc`: the identity token is now scoped with the kernel `boot_id` (`bootid:ticks`), so cross-boot collisions are impossible; unreadable `boot_id` falls back to the previous within-boot token |
| Mixed-version stale reclaim: a pre-v2 build can rename over a just-acquired v2 lock when both race over the same stale lock | P1 | **Documented as accepted residual** (code comment + plan doc). Any representation old builds cannot overwrite would also break their ability to read the owner PID, which the format deliberately preserves; the window closes once no pre-v2 sessions remain |
| `.reclaim` mutex abandoned by a crashed reclaimer whose PID gets reused by a long-lived process would strand reclamation forever | P2 | **Fixed** in `ed595dc`: the mutex now records the reclaimer's start identity; a live PID with a provably different identity is treated as crashed. Unknown identity still always means "assume alive" |
| Race-harness gate used a fixed 1.5 s sleep, so slow contenders could start after the gate opened, weakening the race | P2 | **Fixed** in `ed595dc`: readiness barrier (every contender writes a ready marker before spinning on the gate) plus child cleanup in `finally` |
| No cross-process E2E test of the full replace lifecycle (`admin.shutdown` → runtime stop → lock release → successor acquires) | P2 | **Deferred to a follow-up issue.** `requestBrokerShutdown` and its auth paths are already tested against a real `BrokerSocketServer`, and `replaceBrokerOwner`'s state machine is covered with injected deps; the remaining gap is the composition, which was exercised manually during development |
| Final fence re-check → `SIGTERM` is not atomic (PID reuse in the window) | P3 | **Documented as accepted residual**: closing it needs an identity-bound kernel handle (`pidfd_send_signal`), unavailable without native code under the zero-dependency rule |

Also noted and documented: macOS `ps lstart` has one-second resolution, so same-second PID reuse is indistinguishable — accepted.

Unrelated commit `9b264d5` makes the `model-aware-compaction` disabled-wiring test hermetic (it previously depended on the developer machine's global `~/.pi/agent/settings.json`), which was blocking the pre-push hook on machines with the extension enabled.

Full suite green locally: lint, typecheck, 11/11 packages, broker-core 205 tests, slack-bridge 1625 tests.

Issue #941, opened 2026-07-15
Pinet brokers can become unusable when compaction leaves orphaned tool results


tmustier:
## Summary

Two long-lived Pinet sessions on 11–12 July 2026 became unable to process any further Pinet or Slack work after compaction left tool-call history structurally inconsistent. The broker process remained alive, but every subsequent model turn failed before useful work could resume.

This is a cross-boundary failure involving Pi session history, compaction, extension tool completion, and Pinet’s reliance on a long-lived broker session.

## Observed pattern

- After compaction and a model/session transition, subsequent turns failed because a tool result remained on the active branch without its matching assistant tool call.
- Reloading and compacting again did not repair the history.
- Recovery required stopping the session, backing up the JSONL, reconstructing the missing call/parent relationship, and then resuming.
- A second long-lived worker could still receive Pinet pointers after compaction, but every attempted response failed with an orphaned tool-result/tool-use history error.

In the confirmed case, the orphan involved an extension tool result completing around a compaction boundary. Exact Pi/Pinet versions were not captured by the older logger; the affected broker used `openai-codex/gpt-5.6-sol`.

## Impact

A healthy broker transport and process can appear online while the model session is permanently unable to act on messages. This strands coordination until an operator performs manual session repair or reroutes work to another broker.

## Diagnostic interpretation

The serialized active branch violated the invariant that every tool result must retain its matching assistant tool call. The available evidence does not identify a single owner of the fault. Plausible boundaries include:

- compaction selecting a cut point while an extension tool call is in flight;
- an extension appending its result after the branch was compacted;
- session parent-link reconstruction omitting one side of the call/result pair;
- provider serialization rejecting history that another layer allowed to persist.

Pinet is probably not the origin of the malformed history, but its long-lived broker model turns the corruption into a control-plane outage rather than a disposable session failure.

## Related history

#843 and #595 cover proactive/model-aware compaction, while #915 covers a broker that is alive but unable to consume inbox work for a different reason. This report is specifically about structurally orphaned tool results after compaction.

## Current-status boundary

The evidence is strong for the failure shape but incomplete on component ownership and current reproducibility. No raw session content is included here; a sanitized structural fixture would be required for a focused regression investigation.

tmustier:
## Root cause found

[...]

PR #954, opened 2026-07-28
fix(slack-bridge): hold inbound deliveries while Pi compaction is running

tmustier:
Fixes the delivery side of #941 (see [full RCA](https://github.com/gugu91/pinet/issues/941#issuecomment-5107798487)).

## Problem

Pi's manual/extension compaction path disconnects session persistence while `ctx.isIdle()` still returns `true`. Slack-bridge can therefore inject work during compaction that is never persisted, leaving orphaned tool results and eventually bricking the session.

## Fix

A small compaction gate now blocks every slack-bridge delivery path between `session_before_compact` and `session_compact`:

- queued inbox and maintenance work stays in its existing durable queue
- steering delivery returns `false`, so broker/follower/subtree callers do not acknowledge it before successful injection
- release is deferred one macrotask because `session_compact` fires before Pi reconnects persistence
- abort releases the gate; session start/shutdown resets it
- subtree mode explicitly retries its durable inbox after release and recovers targeted backlog after restart

The gate fails closed if Pi emits neither success nor abort. This pauses inbound delivery until session restart rather than risking session corruption. A timeout cannot safely prove persistence has reconnected.

## Known upstream gap

`session_before_compact` fires after manual compaction disconnects persistence, leaving a small pre-event window. Closing that fully requires a Pi lifecycle signal covering the entire disconnected interval.

## Testing

- top-level regression test proves Slack work is not injected until compaction finishes
- gate tests cover fail-closed delivery, deferred release, abort, and session reset
- subtree tests cover durable retry and restart recovery
- mutation checks killed delivery-bypass, missing-abort, stale-release, missing-retry, and missing-recovery mutants
- `pnpm prepush`: clean
- slack-bridge: 1630 passed, 3 skipped

### DIFF
--- a/slack-bridge/compaction-gate.test.ts
+++ b/slack-bridge/compaction-gate.test.ts
@@ -1,251 +1,57 @@
-import { describe, expect, it, vi } from "vitest";
-import {
-  createCompactionGate,
-  type CompactionGateDeps,
-  type CompactionGateReleaseReason,
-  type CompactionGateTimerHandle,
-} from "./compaction-gate.js";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { createCompactionGate } from "./compaction-gate.js";
 
-interface Harness {
-  gate: ReturnType<typeof createCompactionGate>;
-  onRelease: ReturnType<typeof vi.fn>;
-  /** Run all deferred release callbacks scheduled so far. */
-  flushScheduled: () => void;
-  /** Fire the pending failsafe timer, if armed. */
-  fireFailsafe: () => void;
-  clearedTimers: CompactionGateTimerHandle[];
-}
-
-function createHarness(overrides: Partial<CompactionGateDeps> = {}): Harness {
-  const onRelease = vi.fn<(held: string[], reason: CompactionGateReleaseReason) => void>();
-  const scheduled: Array<() => void> = [];
-  const timers = new Map<number, () => void>();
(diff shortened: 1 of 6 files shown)


tmustier (commented):
Found two blocking delivery-safety issues:

1. **Acknowledged steering messages can be lost permanently** (`slack-bridge/index.ts:309-312`). `holdMessage()` returns `true`, so broker/follower/subtree callers immediately mark the persistent inbox row delivered. The only remaining copy is then in memory. `session_start`/`session_shutdown` explicitly discard it, and `onRelease` also drops it if `sendUserMessage` throws. Please keep the broker/follower inbox as the source of truth until injection succeeds (or carry IDs and acknowledge only after successful injection), and cover shutdown/reload and release-failure paths.

2. **The timeout can recreate the exact corruption this gate is meant to prevent** (`slack-bridge/compaction-gate.ts:120-122`). A manual compaction that is still legitimately running after 10 minutes still has persistence disconnected, but the timeout clears `isCompacting` and calls `onRelease`, which injects held messages. Elapsed time is not evidence that persistence has reconnected. The timeout needs to fail closed (retain/requeue work without delivery) unless an authoritative lifecycle signal confirms the disconnected window ended.

The PR body already documents the smaller pre-event gap; that still needs an upstream Pi fix but is not an additional surprise in this review.

Checks: `pnpm --filter slack-bridge test` (1640 passed, 3 skipped), typecheck, and lint all pass.

PR #964, opened 2026-07-29
fix(slack-bridge): serialize subtree broker startup and return durable spawn handles

### LINKED ISSUE #943: Pinet subtree spawn has non-atomic broker startup and ambiguous registration
## Summary

Pinet subtree spawning showed two repeated failure modes across nine observations from 10–14 July 2026:

1. parallel spawn requests raced on singleton broker startup, so one child started while sibling requests failed;
2. a child process started but did not register within the dispatcher timeout, leaving its ownership and recovery state ambiguous.

The evidence indicates a lifecycle-boundary problem rather than a single slow startup.

## Observed pattern

- Parallel independent spawn calls consistently produced one successful child and one or more `Another pinet broker is already running` failures.
- Repeating the same launches sequentially succeeded.
- Several spawns started a terminal-backed child but timed out after roughly 45 seconds before mesh registration.
- Timeout responses did not provide a stable child/recovery handle, requiring manual process or terminal inspection.
- The typed spawn action could not initialize Pinet from an ordinary non-Pinet session; failure occurred only after the caller had already constructed the delegation payload.

The older papercut logger did not capture exact package versions. These observations occurred across multiple repositories and tasks, primarily with `openai-codex/gpt-5.6-sol` on macOS.

## Impact

Supervisors cannot reliably tell whether a failed spawn means no process exists, a child exists but is late to register, or a sibling launch lost a broker-startup race. This complicates retry and cleanup and can leave unmanaged children behind.

## Diagnostic interpretation

The reports point to two lifecycle phases being coupled behind one synchronous result:

- process/runtime creation and singleton broker acquisition;
- child registration and visibility in the mesh.

Parallel callers appear to contend directly on a singleton startup boundary rather than entering an atomic or serialized spawn state. Registration timeout then collapses “child absent” and “child running but not registered yet” into the same caller-visible failure. The dependency on an already-initialized parent Pinet session is another expression of this implicit lifecycle state.

## Related history

Related closed issues include #761 (sub-worker trees), #834 (spawned worker visibility), and #748 (duplicate-broker startup behavior). Open issue #936 covers ordinary sessions attempting global singleton-broker startup; this report is narrower to subtree spawn concurrency and registration ambiguity.

## Current-status boundary

The evidence postdates the July closure of #761 and #834, but it has not been reproduced against the latest commit. Current behavior should be checked separately for process creation, broker acquisition, and child registration.

tmustier:
Fixes #943

## Acceptance criteria

1. **Single-flight broker startup:** `ensureSubtreeBroker()` shares one in-flight `start()` promise across cold concurrent spawns. The lifecycle test runs three `spawnWorker()` calls through `Promise.all`, asserts three distinct successes and one broker initialization.
2. **Durable timeout handle:** registration timeouts now throw `SubtreeSpawnRegistrationTimeoutError` with `{ launchId, tmuxSessionName, socketPath, state: "launched_unregistered" }`. The dispatcher exposes the handle in both error text and structured details; cleanup targets only the handle's tmux session. Runtime and tool tests cover both surfaces.
3. **Deterministic late registration — adopt:** roster/status reads adopt a late child by matching its broker registration metadata `launchId` to the timed-out launch record. Adoption preserves a child that reached the authoritative broker instead of destructively killing useful work solely because an observer timed out. A delayed-registration test asserts the child appears in the roster and updates the spawn record.
4. **Safe timeout retry:** `retry_handle` is forwarded to the runtime, which validates and cleans that launch before constructing and starting the replacement. The retry test asserts the original session is killed and exactly one child session remains live.
5. **Cheap precondition first:** `runPinetSpawnAction()` checks Pinet, follower role, active extension context, and launcher availability before reading or formatting the delegation payload. The test uses a throwing payload getter and asserts it is never touched in an uninitialized runtime.

## Verification

- `pnpm lint` — passed (all 11 packages plus agent standards)
- `pnpm typecheck` — passed (all 11 packages)
- `pnpm -C slack-bridge exec vitest run --config ../vitest.config.ts subtree-broker-runtime.test.ts pinet-tools.test.ts` — 2 files, 80 tests passed
- `pnpm -C slack-bridge test` — 92 files passed, 2 skipped; 1,639 tests passed, 3 skipped
- `pnpm test` — passed across the workspace (run by the pre-push hook)

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -12,6 +12,10 @@
 ## Design philosophy
 
 - Guiding principle for this change:
+- Simplicity review:
+  - [ ] Comments explain non-obvious reasons or invariants; clearer code replaces narration.
+  - [ ] Tests exercise states reachable through production boundaries rather than bypassing them.
+  - [ ] One-use helpers are inlined unless they form a documented semantic seam.
 - If this touches Pinet / `slack-bridge`: how does it preserve token-efficient progressive discovery?
   - Hot-path schemas/prompts stay compact:
   - Cold paths remain discoverable through dispatcher `help` / per-action schemas or docs/skills:
--- a/scripts/agent-standards-lint.mjs
+++ b/scripts/agent-standards-lint.mjs
@@ -164,93 +164,153 @@ function isExportedNode(node) {
   return Boolean(node.modifiers?.some((modifier) => modifier.kind === ts.SyntaxKind.ExportKeyword));
 }
 
+function isExportedVariableDeclaration(node) {
+  return ts.isVariableDeclarationList(node.parent) && isExportedNode(node.parent.parent);
+}
+
 function hasSingleUseHelperIgnore(sourceText, position) {
   const precedingText = sourceText.slice(Math.max(0, position - 300), position);
   return /agent-standards-ignore\s+prefer-inline-single-use-helper/.test(precedingText);
 }
 
-function collectTopLevelHelperNames(sourceFile) {
-  const helperNames = new Set();
-
-  for (const statement of sourceFile.statements) {
-    if (ts.isFunctionDeclaration(statement) && statement.name && !isExportedNode(statement)) {
-      helperNames.add(statement.name.text);
-      continue;
-    }
-
-    if (!ts.isVariableStatement(statement) || isExportedNode(statement)) continue;
-    for (const declaration of statement.declarationList.declarations) {
-      if (!ts.isIdentifier(declaration.name)) continue;
-      const initializer = declaration.initializer;
-      if (
-        initializer &&
-        (ts.isArrowFunction(initializer) || ts.isFunctionExpression(initializer))
-      ) {
-        helperNames.add(declaration.name.text);
-      }
-    }
-  }
-
-  return helperNames;
+function collectExportedSymbols(sourceFile, checker) {
+  const moduleSymbol = checker.getSymbolAtLocation(sourceFile);
+  if (!moduleSymbol) return new Set();
+  return new Set(
+    checker
+      .getExportsOfModule(moduleSymbol)
+      .map((symbol) =>
+        symbol.flags & ts.SymbolFlags.Alias ? checker.getAliasedSymbol(symbol) : symbol,
+      ),
+  );
 }
 
-function collectTopLevelHelperDeclarations(
+function isLocalHelperDeclaration(node, exportedSymbols, checker) {
+  if (ts.isFunctionDeclaration(node)) {
+    const symbol = node.name ? checker.getSymbolAtLocation(node.name) : undefined;
+    return Boolean(node.name && !isExportedNode(node) && symbol && !exportedSymbols.has(symbol));
+  }
+  if (
+    !ts.isVariableDeclaration(node) ||
+    !ts.isIdentifier(node.name) ||
+    isExportedVariableDeclaration(node)
+  ) {
+    return false;
+  }
+  const symbol = checker.getSymbolAtLocation(node.name);
+  if (!symbol || exportedSymbols.has(symbol)) {
+    return false;
+  }
+  return Boolean(
+    node.initializer &&
+    (ts.isArrowFunction(node.initializer) || ts.isFunctionExpression(node.initializer)),
+  );
+}
+
+function helperKey(node) {
+  const names = [node.name.text];
+  for (let parent = node.parent; parent; parent = parent.parent) {
+    if ((ts.isFunctionDeclaration(parent) || ts.isFunctionExpression(parent)) && parent.name) {
+      names.push(parent.name.text);
+    } else if (
+      (ts.isFunctionExpression(parent) || ts.isArrowFunction(parent)) &&
+      ts.isVariableDeclaration(parent.parent) &&
+      ts.isIdentifier(parent.parent.name)
+    ) {
+      names.push(parent.parent.name.text);
+    } else if (ts.isMethodDeclaration(parent) && ts.isIdentifier(parent.name)) {
+      names.push(parent.name.text);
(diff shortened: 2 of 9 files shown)


tmustier:
## Pre-merge review (gpt-5.6-sol xhigh): REQUEST-CHANGES

MERGE VERDICT: REQUEST-CHANGES

## Blocking issues

1. **Failed startup can leave a live, unreachable broker and make every retry conflict** — `slack-bridge/subtree-broker-runtime.ts:777-848,861-868`. The in-flight promise is cleared correctly, but once `startBroker()` succeeds, failures in `getAgentMetadata()`, self-registration, or subsequent initialization do not stop the local broker. `activeBroker` may still be null, so retry reacquires against the leaked lock/socket and fails. **Fix:** wrap post-`startBroker()` initialization in rollback that `await broker.stop()` before rethrowing; publish active state only after initialization is coherent. Add failure-after-listen→successful-retry coverage.

2. **Cleanup can kill a different tmux session and accepts an unauthenticated target when the launch record is absent** — `slack-bridge/subtree-broker-runtime.ts:666-685`. Bare `-t sessionName` permits tmux prefix matching (tmux documents `=name` for exact matching); after the intended session disappears, a prefix-related session can be killed. If `spawnedWorkers.get(launchId)` is absent, a caller-controlled handle with the current Pinet socket can name any session, and cleanup recomputes rather than carries the launch-time tmux socket. **Fix:** use an immutable exact tmux target (`-t =name`) on the launch-time tmux socket and validate the handle against a live/persisted launch record (or authenticate/sign it); reject absent/mismatched records unless a durable persisted mapping proves ownership.

3. **Retry is not fenced against a registration already in flight, so it can still leave the old child beside the replacement** — `slack-bridge/subtree-broker-runtime.ts:673-700,881-912`. Cleanup performs one DB scan and then deletes the record, but there is no cleaned-launch tombstone. The old child can register after that scan even though its tmux session was killed, while retry launches a second child. **Fix:** mark the old launchId fenced/consumed before cleanup, reject or immediately unregister every future registration carrying it, and launch the replacement only after that fence is active. Test a registration delivered during/after cleanup.

4. **Cleanup suppresses operational tmux failures as if the session were already absent** — `slack-bridge/subtree-broker-runtime.ts:666-685`. Every `has-session` rejection is converted to `Error("missing")`, which `cleanupSpawn()` ignores. A socket, permission, or transport failure can therefore leave the old process alive while relaunch proceeds. **Fix:** preserve the real failure and suppress only a positively identified missing exact target; test that a non-missing probe/kill failure prevents `new-session`.

## Non-blocking nits

- `spawnWorker()` now starts the broker before rejecting empty task/repo (`subtree-broker-runtime.ts:875-880`); move cheap input validation back before startup.
- Public `runtime.start()` remains outside the single-flight promise, so `/pinet subtree start` can still race an auto-start (`index.ts:1662`); make the public start boundary single-flight.
- Late “adoption” updates only roster metadata; because task delivery happens after the wait (`subtree-broker-runtime.ts:925-978`), a post-timeout child never receives its original task. Represent that as unassigned/cleanup-required or deliver pending work exactly once.
- The slash command cannot provide the new retry handle, so repeating `/pinet subtree spawn` can duplicate a late child; expose recovery there or clearly refuse unsafe retries.
- Replace the 5/15/25 ms real-timer adoption test with fake timers or an explicit deferred registration gate.

## Test-gap list

- Post-listen startup failure followed by successful retry.
- Direct `start()` racing cold spawn.
- Real tmux exact-target/prefix behavior and launch-time tmux socket use.
- Non-missing tmux cleanup error blocks relaunch.
- Registration racing cleanup is fenced and leaves one live child/no old DB row.
- Repeated/concurrent retry-handle consumption is idempotent.
- Adopted late child receives or is explicitly marked as not having received the original task.
- Stop racing timeout preserves all immutable handle fields.
- Slash-command timeout recovery/no-duplicate behavior.

## Rationale

The PR fixes the uncomplicated cold-start race, returns useful structured timeout data, and adds meaningful happy-path lifecycle tests. It is not safe to merge yet because failure rollback is incomplete and the cleanup/retry contract can target the wrong tmux session, suppress a failed cleanup, or admit a late old registration alongside the replacement—the exact orphan/double-launch class the acceptance criteria require eliminating.

tmustier:
## Re-review (gpt-5.6-sol xhigh): APPROVE-WITH-NITS

All four original blockers plus the residual retry-handle blocker are now **CLOSED**, verified against the fix commits (`4f657d5`…`2045bfb`):

1. ✅ Startup rollback — post-`startBroker()` init failures stop the broker before rethrowing; retry succeeds.
2. ✅ Tmux cleanup — exact-target (`=name`) on the launch-time socket; unrecorded handles rejected; fake harness now models prefix matching.
3. ✅ Fencing — fence installed before kill/delete; registration resolver rejects fenced launchIds; retry handles are permanently consumed once a replacement launch is attempted (consumption installed immediately before `new-session`, so a rejected command can't be mistaken for non-launch); second retry with a consumed handle returns the terminal outcome carrying the replacement's handle instead of double-launching.
4. ✅ Operational tmux failures propagate — only a positively identified missing exact target is suppressed.

Nits fixed along the way: input validation before broker startup, single-flight public `start()`, real socket-registration fencing test, launch-time Pinet socket captured in timeout handles, env-mutation test proving launch-time tmux socket use.

Verification: 1,649 slack-bridge tests passed (3 skipped), typecheck/lint/prettier clean; revert-mutation probes confirmed the new tests fail if the fixes are reverted.

Residual non-blocking nits (fine to defer):
- Pre-launch cleanup retryability is implemented but not fully covered by a checked-in test.
- `fencedLaunchIds` / consumed `retrySpawns` entries grow unbounded until runtime stop — acceptable for low-volume subtree retries; consider a compact durable consumed-state record if volume grows.

Issue #962, opened 2026-07-29
design: hibernation for supervised subtree children — lift supervised_subtree_unsupported, define wake ordering and orphan policy


tmustier:
## Current state (deliberate scoping, now a gap)

`evaluateHibernateEligibility` explicitly refuses supervised subtree children (`broker-core/lifecycle.ts:~43-48`):

```ts
if (agent.parentAgentId || agent.supervisionState === "supervised") {
  return { eligible: false, reason: "supervised_subtree_unsupported" };
}
```

So subtree children can never sleep. Combined with the subtree broker living inside the parent worker process, the durability story for children is unfinished by design:

- a hibernated or dead parent leaves children with **no mailbox endpoint** (the subtree SQLite DB persists under `~/.pi/pinet-subtrees/<worker>/`, but nothing serves it) and **no supervisor**;
- children burn resources indefinitely because nothing may checkpoint them.

This is supporting evidence that child durability remains open — note that daemon-owned spawn (#961) does **not** by itself solve wake ordering or orphan policy; this issue owns those decisions.

## Design questions to resolve

1. **Checkpoint ordering** — may a supervising parent hibernate while children are live? Today the parent (no `parentAgentId`) can pass eligibility while its in-process subtree broker would vanish, stranding children. Enumerate and close this hazard.
2. **Wake ordering** — child wake requires a live mailbox endpoint (parent subtree broker, or daemon under #961). Who wakes the parent when a sleeping child needs delivery? Parent-before-child as an invariant, or endpoint-independent children?
3. **Orphan ownership** — who supervises children when the parent is gone, distinguishing parent-crash from parent-hibernate.
4. **Identity and fencing** — extend `stableId` + generation-fence reservations (#923/#924 primitives) to children, so a woken child cannot race a stale sibling generation.
5. **Topology dependency** — the isolated-subtree-DB vs central-registration decision in #961 determines which supervisor can even see children. State the dependency explicitly; do not begin implementation before #961 resolves it.

## Acceptance shape (design phase)

A written design that answers 1–5 with lifecycle-state diagrams, extends the eligibility matrix (replacing the blanket `supervised_subtree_unsupported` with reasoned per-case verdicts), and defines E2E scenarios: parent-hibernate-with-live-children, child-hibernate-and-wake, parent-crash-orphan-recovery.

## References

#923, #924, #961, #957, `plans/hibernation-live-activation.md`, `plans/761-pinet-worker-subtrees.md`.

Issue #958, opened 2026-07-29
feat: let workers run on Herdr, not just tmux (staged migration)


tmustier:
## What this is about

Pinet assumes every worker runs inside tmux. That assumption isn't in one place — it's baked into the database schema, the types, and the spawn/hibernation code. We want workers to be able to run in Herdr instead (Herdr by default, tmux as fallback), without breaking anything that works today.

Because tmux is baked into *stored data*, this is a data-migration project first and an "add a second backend" project second.

## Can Herdr do what we need? ✅ Verified yes (Herdr 0.7.5, 2026-07-29)

- **Relaunch into the same pane** after a process dies — works.
- **Env vars at launch** — works: create the pane with `--env`. (You can't pass env to `pane run` because that just types the command into the shell — same as tmux send-keys. Env belongs to pane creation, where Herdr actually spawns the shell.)
- **Knowing exactly which process we launched** — works: Herdr owns the pane's shell PID, and our workers already tell the broker their own PID when they register.
- **Killing a stuck launch** without touching its neighbours — works.
- **Looking at a dead worker's pane afterwards** — works: `herdr terminal attach <terminal-id>`, plus a headless observe mode. (A worker dying returns the pane to its shell with scrollback intact; only the shell dying closes the pane.)
- **Server restart** — pane layout and IDs survive; processes die, which we already assume after any restart. Scrollback survives too if `pane_history` is turned on in config (verified live).

Full write-ups: [spike report](https://github.com/tmustier/pi-dial/blob/main/docs/herdr-spike-958-report.md) · [source-level stock-take](https://github.com/tmustier/pi-dial/blob/main/docs/herdr-upstream-asks.md). No Herdr changes or upstream requests needed.

Three things the implementation must remember:

1. Herdr's `terminal_id` changes on every server restart — always look it up fresh, never store it as identity.
2. Herdr declined pane-level hibernation upstream, so #962 owns that layer entirely.
3. Turn on `pane_history` for pinet-owned Herdr sessions via a dedicated `XDG_CONFIG_HOME` (it's off by default upstream because scrollback can contain secrets).

### PR A — prep, zero behaviour change

- Schema: add a runtime-kind discriminant column to runtime rows (NOT NULL, default `tmux`, CHECK-constrained to the supported kinds). The tmux columns stay NOT NULL for now — no speculative nullability before a second backend exists. PR B rebuilds the table when the widened union and its payload shape are real.
[...]

PR #969, opened 2026-07-29
feat(broker): runtime-kind schema and worker launch seam (tmux-only)

tmustier:
PR A from #958: make the stored runtime model runtime-neutral and put the worker spawn path behind a launch seam, with tmux as the only backend. Zero behaviour change.

## What's in here

**Part 1 — schema + types** (`ed53c34`, `1a97961`, `226af27`)
- `AgentRuntimeSpec` gains a required `runtimeKind: "tmux"` discriminant (persisted and read back). Chose a required literal field over a full discriminated union to keep call-site ripple small; PR B widens it to a union when a second kind exists. `RuntimeAttemptHandle` is unchanged — handles aren't persisted, so the recorded-backend contract lives on the spec.
- Broker schema v22 → v23: `agent_runtime_specs` gains `runtime_kind TEXT NOT NULL DEFAULT 'tmux'` with `CHECK(runtime_kind IN ('tmux'))` via a single `ADD COLUMN` (no rebuild). The three tmux columns stay NOT NULL — the schema admits exactly the states the types can represent. PR B rebuilds the table when the widened union and backend payload shape actually exist. Existing rows migrate as `runtime_kind = 'tmux'`.
- Tests: fresh-DB and v22→v23 migration coverage (rows read back as tmux).

**Part 2 — launch seam** (`5098e96`)

_Post-review (`606029728`, `5fb64cc`): migration test asserts all 20 legacy columns + PK; nullable rebuild reverted to plain ADD COLUMN; `RuntimeAttemptHandle` discriminant and public `probe` removed per inline review._
- `WorkerRuntimeController` (plain object: `launch` / `cleanup`, with session probing private to the tmux cleanup) extracted in `slack-bridge/subtree-broker-runtime.ts`, tmux as the only implementation, selected by the spec's **recorded** `runtimeKind` so probe/cleanup always dispatch on what the worker was launched with.
- The #964 hardening (launch fences, retry-handle consumption immediately before launch, socket-authenticated exact-name cleanup, operational-failure propagation) is moved, not changed.

## tmux touchpoint checklist (grep-backed)

Behind the seam (this PR):
- [x] `slack-bridge/subtree-broker-runtime.ts` — spawn-path launch / probe / socket-authenticated cleanup

PR B (must become runtime-aware when the Herdr backend lands):
- [ ] `subtree-broker-runtime.ts` broker-stop teardown — `childTmuxSessions()` enumeration + hardcoded `controllers.tmux.cleanup(...)`; must enumerate specs by recorded kind
- [ ] `slack-bridge/pinet-tools.ts` — spawn/monitor hint strings reference tmux sessions
- [ ] `slack-bridge/runtime-mode.ts`, `slack-bridge/runtime-agent-context.ts` — `PINET_LAUNCH_SOURCE = "broker-tmux"` launch-source constant needs a herdr value
- [ ] `slack-bridge/broker/socket-server.ts` — registration params carry `tmuxSession`; needs a neutral locator
- [ ] `slack-bridge/pinet-session-formatting.ts` — session display reads `tmuxSession`
- [ ] `broker-core/lifecycle.ts` — `tmuxSession` field reference
- [ ] `slack-bridge/prompts/broker/tmux.md` + `broker-prompt-loader.ts` — broker prompt instructs tmux launches; loader defaults to `tmux.md`

Out of scope (reason):
- Hibernation family (`broker-core/hibernation-orchestrator.ts`, `hibernation-status.ts`, `slack-bridge/broker/hibernation-*`, `slack-bridge/helpers.ts` comment) — #958 explicitly decouples hibernation; #962 owns hibernate/wake on Herdr
- `broker-core/types.ts` / `broker-core/schema.ts` — migrated by this PR
- `nvim-bridge/comments.ts` (`detectTmuxLabel`) — labels the human's own tmux environment; unrelated to worker runtime
- `plans/*.md` — historical planning docs, not runtime surfaces

## Verification

- Slack bridge: typecheck clean; 92 files passed / 2 skipped, **1,653 tests passed / 3 skipped**
- broker-core: 207 tests passed; repo-wide lint + typecheck clean (11 packages)
- Migration tests fail if the discriminant or v23 migration is reverted

refs #958

### DIFF
--- a/broker-core/hibernation-orchestrator.ts
+++ b/broker-core/hibernation-orchestrator.ts
@@ -47,6 +47,8 @@ export interface RuntimeLaunchContext {
  * (whose recorded PID generation is already dead) nor a different attempt.
  */
 export interface RuntimeAttemptHandle {
+  /** Runtime backend discriminant. Only tmux is supported until PR B. */
+  readonly runtimeKind: "tmux";
   /** The reservation nonce of the launching attempt (binds the handle to it). */
   readonly reservationNonce: string;
   /** The tmux pane the attempt launched into (a real adapter locates the PID). */
--- a/broker-core/hibernation-orchestrator.ts
+++ b/broker-core/hibernation-orchestrator.ts
@@ -47,8 +47,6 @@ export interface RuntimeLaunchContext {
  * (whose recorded PID generation is already dead) nor a different attempt.
  */
 export interface RuntimeAttemptHandle {
-  /** Runtime backend discriminant. Only tmux is supported until PR B. */
-  readonly runtimeKind: "tmux";
   /** The reservation nonce of the launching attempt (binds the handle to it). */
   readonly reservationNonce: string;
   /** The tmux pane the attempt launched into (a real adapter locates the PID). */
--- a/broker-core/schema.ts
+++ b/broker-core/schema.ts
@@ -948,7 +948,7 @@ export function defaultDbPath(): string {
 
 export const DEFAULT_RESUMABLE_WINDOW_MS = 15_000;
 export const DEFAULT_DISCONNECTED_PURGE_GRACE_MS = 60 * 60_000;
-export const CURRENT_BROKER_SCHEMA_VERSION = 22;
+export const CURRENT_BROKER_SCHEMA_VERSION = 23;
 
 /**
  * Lifecycle states whose durable identity, inbox, thread ownership, and runtime
@@ -1729,7 +1729,8 @@ function createAgentHibernationTables(db: DatabaseSync): void {
     CREATE TABLE IF NOT EXISTS agent_runtime_specs (
       agent_id TEXT PRIMARY KEY NOT NULL, stable_id TEXT NOT NULL, broker_owner_id TEXT NOT NULL,
       cwd TEXT NOT NULL, repo_root TEXT NOT NULL, worktree_path TEXT NOT NULL,
-      tmux_socket TEXT NOT NULL, tmux_session TEXT NOT NULL, tmux_target TEXT NOT NULL,
+      runtime_kind TEXT NOT NULL DEFAULT 'tmux',
+      tmux_socket TEXT, tmux_session TEXT, tmux_target TEXT,
       executable TEXT NOT NULL, argv_json TEXT NOT NULL, env_allowlist_json TEXT NOT NULL,
       session_resume_ref TEXT NOT NULL, config_fingerprint TEXT NOT NULL,
       expected_host TEXT NOT NULL, expected_user TEXT NOT NULL, launch_source TEXT NOT NULL,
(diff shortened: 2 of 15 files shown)


tmustier:
## Pre-merge review (gpt-5.6-sol xhigh): APPROVE-WITH-NITS

No blockers. Line-by-line check confirms the extraction is a structural move: all #964 hardening invariants (fence-before-cleanup, recorded-kind dispatch, launch-time socket reuse, `-t "=name"` targeting, operational-error propagation, error/hint strings byte-for-byte) are preserved. v22→v23 rebuild copies all 20 legacy columns explicitly, recreates the PK, backfills `runtime_kind='tmux'`, and follows the per-version transaction pattern. Exactly two direct `.tmux` dispatches exist: launch selection (necessarily tmux in a tmux-only PR) and the documented broker-stop teardown exception staged for PR B.

Nit (being addressed on this branch): the v22→v23 migration test asserts only the discriminant + tmux fields — expanding it to assert all legacy columns and the recreated PK constraint so a wrong copy mapping can't escape.

Verification: 95 focused tests + full slack-bridge suite (1,653 passed / 3 skipped), broker-core 207, repo lint/typecheck clean.

broker-core/schema.ts
```diff
@@ -1857,6 +1858,43 @@ function addRuntimeSpecVcsIdentityColumn(db: DatabaseSync): void {
...
+// switch a readable index.
+function addRuntimeSpecKindColumn(db: DatabaseSync): void {
+  if (getTableColumns(db, "agent_runtime_specs").has("runtime_kind")) return;
+  db.exec(`
+    ALTER TABLE agent_runtime_specs RENAME TO agent_runtime_specs_legacy;
+    CREATE TABLE agent_runtime_specs (
+      agent_id TEXT PRIMARY KEY NOT NULL, stable_id TEXT NOT NULL, broker_owner_id TEXT NOT NULL,
+      cwd TEXT NOT NULL, repo_root TEXT NOT NULL, worktree_path TEXT NOT NULL,
+      runtime_kind TEXT NOT NULL DEFAULT 'tmux',
+      tmux_socket TEXT, tmux_session TEXT, tmux_target TEXT,
```
tmustier: **Blocking:** this weakens the live invariant before another backend exists. `AgentRuntimeSpec` and `getAgentRuntimeSpec()` still claim all three tmux locators are strings and `runtimeKind` is always `"tmux"`; no parser or alternate dispatch can represent rows admitted by this schema. The fresh-DB test then locks in that impossible current state. Keep the tmux columns `NOT NULL` and make v23 a simple `ADD COLUMN runtime_kind TEXT NOT NULL DEFAULT 'tmux'` (ideally constrained to the supported literal). PR B can rebuild the table when it introduces the discriminated union and actually needs nullable backend-specific payloads. This removes speculative flexibility, the rebuild, and the nullability test together.

broker-core/hibernation-orchestrator.ts
```diff
@@ -47,6 +47,8 @@ export interface RuntimeLaunchContext {
  * (whose recorded PID generation is already dead) nor a different attempt.
  */
 export interface RuntimeAttemptHandle {
+  /** Runtime backend discriminant. Only tmux is supported until PR B. */
+  readonly runtimeKind: "tmux";
```
tmustier: This discriminant is never read, while the same handle still requires tmux-specific `tmuxTarget`. It therefore adds fixture/type churn without creating a usable runtime-neutral contract. Please defer it until the handle becomes a real discriminated union and dispatch consumes the field.

slack-bridge/subtree-broker-runtime.ts
```diff
@@ -56,10 +57,25 @@ export interface SubtreeWorkerRecord {
...
+  runtimeKind: AgentRuntimeSpec["runtimeKind"];
+  sessionName: string;
+  tmuxSocketPath: string | null;
+}
+
+interface WorkerRuntimeController {
+  createLaunchSpec: (sessionName: string) => WorkerRuntimeSpec;
+  monitorCommand: (spec: WorkerRuntimeSpec) => string;
+  launch: (spec: WorkerRuntimeSpec, launcherPath: string) => Promise<void>;
+  probe: (spec: WorkerRuntimeSpec) => Promise<boolean>;
```
tmustier: `probe` is not part of the controller contract in practice: no caller invokes it; only this controller's `cleanup` closure does. Keeping it public creates dead surface and requires a lint-waived one-use helper. Remove it from the interface/returned object and keep the check private to `cleanup` (inline it unless a real second call site appears).

tmustier (commented):
Focused review for excessive defensiveness/verbosity: one blocking schema/type mismatch and two smaller staging abstractions. The v22 migration's full-row assertion is worth keeping because a table rebuild can genuinely mis-map legacy data; the fresh-schema nullability assertion is the implausible-state test.

broker-core/schema.ts
tmustier: Accepted — fixed in 5fb64cc. v23 is now a single `ADD COLUMN runtime_kind TEXT NOT NULL DEFAULT 'tmux' CHECK(runtime_kind IN ('tmux'))` (CHECK matches existing status/operation column conventions); the three tmux columns stay NOT NULL and the rebuild is reverted, deferred to PR B when the union and backend payload shape actually exist. Fresh-DB test now asserts the columns remain NOT NULL; the full-row migration assertion and PK check are kept as you suggested.

broker-core/hibernation-orchestrator.ts
tmustier: Accepted — fixed in 5fb64cc. `runtimeKind` removed from `RuntimeAttemptHandle` and all fixture churn reverted (adapters, commands, e2e, orchestrator tests). Handles aren't persisted, so the recorded-backend contract lives on `AgentRuntimeSpec`, which keeps its discriminant and persistence threading. Handle gets the real discriminated union in PR B when dispatch consumes it.

slack-bridge/subtree-broker-runtime.ts
tmustier: Accepted — fixed in 5fb64cc. `probe` removed from the interface and returned object; the has-session check is private to the tmux controller's cleanup, and the one-use lint waiver is gone. Interface is now launch/cleanup.

tmustier:
Second-pass review resolved in `44683c3`:

- trimmed the v22→v23 test to the behavior now owned by the additive migration (`runtime_kind` + schema version), removing 20-column and PK assertions that only retested SQLite
- batched the retention test’s 10,005 seed inserts in one transaction after that untouched test timed out in CI twice; behavior assertions are unchanged

Verification: final reviewer found no findings; 5 consecutive focused lifecycle runs passed; repo-wide lint, typecheck, and tests passed (broker-core 207/207, slack-bridge 1,653 passed / 3 skipped). Pre-push gates passed again during push.

PR #971, opened 2026-07-29
Fix Slack Socket Mode reconnect race

### LINKED ISSUE #970: Slack Socket Mode reconnect race leaks errors into the Pi TUI
## Summary

A long-lived Pinet broker can intermittently print the following directly into Pi's interactive TUI while idle:

```text
[slack-adapter] Socket Mode: Sent before connected.
```

The broker remains usable and a subsequent TUI redraw removes the stray lines. This is not session or chat-stream corruption: the message is written directly to stderr by the Slack adapter.

## Root cause

`SlackSocketModeClient` stores the current connection in mutable `this.ws`, but message listeners from every socket call the shared `handleFrame()` method:

- `connectSocketMode()` replaces `this.ws` with a newly created WebSocket.
- A Slack `disconnect` frame schedules reconnection without retiring or isolating the old socket.
- `handleFrame()` ACKs an envelope through `this.ws?.send(...)`, rather than through the socket that delivered the frame.
- If an old socket delivers a late frame while the replacement is still `CONNECTING`, the ACK is sent through the replacement.
- Undici throws `DOMException("Sent before connected.", "InvalidStateError")` for that operation.
- `SlackAdapter` forwards the error through `console.error`, bypassing Pi's differential renderer and temporarily corrupting the visible terminal.

Relevant paths:

- `slack-bridge/slack-access.ts`: `SlackSocketModeClient.connectSocketMode()` and `handleFrame()`
- `slack-bridge/broker/adapters/slack.ts`: Socket Mode `onError` callback

There is a second reconnect hazard in the same lifecycle: a stale socket's later `close` event can schedule another reconnect even when its replacement is healthy, potentially causing repeated connection churn.

## Expected behaviour

- Frames are acknowledged on the WebSocket that delivered them.
- Stale socket events cannot affect the active connection or schedule redundant reconnects.
- Background adapter errors do not write raw output into Pi's interactive TUI.

## Suggested fix

1. Capture each socket in its event-listener closures and pass the originating socket into frame handling.
2. ACK only through that socket and only when its `readyState` is `OPEN`.
3. Guard `open`, `close` and reconnect handling by socket identity; prevent overlapping reconnect attempts.
4. Retire the previous socket once the replacement is established.
5. Replace direct `console.error` reporting with an injected, throttled UI/status or file-logging path.

## Regression coverage

Add tests which prove:

- a late frame from the previous socket never sends through a `CONNECTING` replacement
- an envelope ACK targets its originating socket
- closing a stale socket does not schedule another reconnect
- multiple disconnect/close signals result in only one reconnect attempt
- Socket Mode failures do not write directly to the Pi TUI

tmustier:
## Issue / problem solved

- Closes/Fixes/Refs: Fixes #970
- Why this is prioritized: Slack Socket Mode reconnect races can write errors directly into long-lived interactive Pi TUI lanes.

## Troubleshooting / diagnosis

- How the problem was reproduced or investigated: traced Socket Mode reconnects through `SlackSocketModeClient` and the broker adapter's error callback, then reproduced the stale-socket sequence with a WebSocket fake that enforces real `readyState` send behavior.
- Evidence for the root cause: frame handlers ACKed through mutable `this.ws`, so a late frame from the retired socket could send through its connecting replacement and throw `Sent before connected.`. The broker adapter then emitted that error with `console.error`.
- Relevant constraints or edge cases: events queued by a retired socket must remain bound to that socket; retired close/disconnect events must not schedule another reconnect; connection attempts must not overlap.

## Design philosophy

- Guiding principle for this change: make each socket own its events and keep interactive transport state inside the existing status UI.
- Simplicity review:
  - [x] Comments explain non-obvious reasons or invariants; clearer code replaces narration.
  - [x] Tests exercise states reachable through production boundaries rather than bypassing them.
  - [x] One-use helpers are inlined unless they form a documented semantic seam.
- If this touches Pinet / `slack-bridge`: how does it preserve token-efficient progressive discovery?
  - Hot-path schemas/prompts stay compact: no schema or prompt changes.
  - Cold paths remain discoverable through dispatcher `help` / per-action schemas or docs/skills: unchanged.
  - Templates/examples/large usage guidance stay out of always-loaded prompts/tool schemas: unchanged.

## Repo design fit

- Existing architecture/conventions this follows: extends the existing Socket Mode callback and extension-status paths rather than adding a separate notification or logging system.
- Extension/package boundaries preserved: all changes stay within `slack-bridge`.
- Guardrail, security, auth, or local-power implications: none.

## Alternatives considered

- Alternative(s) considered: capture raw extension stderr in Pi core; notify the user on every transport error; retain mutable-socket ACK routing with a `readyState` guard.
- Why they were not chosen: the race belongs to Pinet, repeated notifications would be noisy, and a guard alone would silently drop ACK ownership rather than fixing it.

## Testing / validation

- Tests added or updated: stale-frame socket ownership, stale-close reconnect suppression, overlapping reconnect prevention and controlled runtime error-status routing.
- Commands run:
  - [x] `pnpm lint`
  - [x] `pnpm typecheck`
  - [x] `pnpm test`
  - [x] Other: `pnpm format:check`, `git diff --check`
- Manual validation: the regression fake throws the same `DOMException("Sent before connected.", "InvalidStateError")` as a real WebSocket when a send targets a connecting socket.

## Review notes / follow-ups

- Known limitations or non-blocking concerns: none identified.
- Follow-up issues/PRs: none.
- Reviewer focus areas: socket identity checks in reconnect lifecycle and whether extension status is the correct broker/single-player error surface.

### DIFF
--- a/slack-bridge/broker/adapters/slack.test.ts
+++ b/slack-bridge/broker/adapters/slack.test.ts
@@ -3510,6 +3510,36 @@ describe("SlackAdapter — e2e Socket Mode lifecycle", () => {
     }
   });
 
+  it("classifies message handler failures as event errors", async () => {
+    const handlerError = new Error("message failed");
+    const onError = vi.fn();
+    const client = new SlackSocketModeClient({
+      slack: vi.fn(async () => ({ url: "wss://slack.test/socket" })),
+      botToken: "xoxb-test",
+      appToken: "xapp-test",
+      resolveBotUserIdOnConnect: false,
+      onMessage: () => {
+        throw handlerError;
+      },
+      onError,
+    });
+
+    await client.connect();
+    const socket = FakeWebSocket.instances[0]!;
+    socket.open();
+    socket.emit("message", {
+      data: JSON.stringify({
+        type: "events_api",
+        payload: { event: { type: "message" } },
+      }),
+    });
+
+    await waitForAssertion(() => {
+      expect(onError).toHaveBeenCalledWith(handlerError, "event");
+    });
+    await client.disconnect();
+  });
+
   it("routes Socket Mode connection failures through the adapter callback", async () => {
     fetchMock.mockImplementation(
       async (input) =>
--- a/slack-bridge/broker/adapters/slack.test.ts
+++ b/slack-bridge/broker/adapters/slack.test.ts
@@ -3320,7 +3320,7 @@ describe("SlackAdapter — e2e Socket Mode lifecycle", () => {
     static instances: FakeWebSocket[] = [];
 
     readonly url: string;
-    readyState = FakeWebSocket.OPEN;
+    readyState = FakeWebSocket.CONNECTING;
     sent: string[] = [];
     private readonly listeners = new Map<string, Array<(event: { data?: string }) => void>>();
 
@@ -3342,6 +3342,11 @@ describe("SlackAdapter — e2e Socket Mode lifecycle", () => {
       if (this.readyState !== FakeWebSocket.CLOSED) {
         this.sent.push(String(data));
       }
+    }
+
+    open(): void {
+      this.readyState = FakeWebSocket.OPEN;
+      this.emit("open", {});
     }
 
     close(): void {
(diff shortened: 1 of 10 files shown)


tmustier (commented):
Two findings before merge:

1. **Do not turn every frame-handler failure into a persistent connection error.**  routes exceptions from , reactions, interactions and slash commands through the same  callback as  failures (). Both runtimes now translate that callback to the transport  state, and  refuses to overwrite it until another socket . A single inbound processing error can therefore leave a healthy connection showing  indefinitely. Separate transport lifecycle errors from handler/delivery errors; preserve diagnostics without raw TUI output.

2. **Keep the originating-socket invariant decisive.** Production always calls  with a real socket, but it now accepts  and optional-chains  because four white-box tests invoke the private method with  (, ). Require , use a minimal open fake in those tests, and check  before ACKing, as issue #970 specifies.

The new / state and socket-identity checks otherwise correspond to reachable reconnect races; I would keep them. The added lifecycle tests exercise plausible sequences rather than hypothetical states.

tmustier (commented):
Correction: shell quoting stripped the code spans from my previous review. The findings with references are:

1. **Do not turn every frame-handler failure into a persistent connection error.** `SlackSocketModeClient` routes exceptions from `onMessage`, reactions, interactions and slash commands through the same `onError` callback as `apps.connections.open` failures (`slack-access.ts:626-650`). Both runtimes now translate that callback to the transport `✗` state, and `updateBadge()` refuses to overwrite it until another socket `open`. A single inbound processing error can therefore leave a healthy connection showing `✗` indefinitely. Separate transport lifecycle errors from handler/delivery errors; preserve diagnostics without raw TUI output.

2. **Keep the originating-socket invariant decisive.** Production always calls `handleFrame` with a real socket, but it now accepts `WebSocket | null` and optional-chains `send` because four white-box tests invoke the private method with `null` (`slack-access.ts:661-669`, `broker/adapters/slack.test.ts:1203,1208,1251,1368`). Require `WebSocket`, use a minimal open fake in those tests, and check `readyState === WebSocket.OPEN` before ACKing, as issue #970 specifies.

The new `connecting`/`pendingSocket` state and socket-identity checks otherwise correspond to reachable reconnect races; I would keep them. The added lifecycle tests exercise plausible sequences rather than hypothetical states.

PR #972, opened 2026-07-29
feat: Herdr launch backend for subtree workers (PR B of #958)

tmustier:
PR B of #958: subtree workers can now launch on Herdr instead of tmux, behind a config setting that defaults to tmux. Zero behavior change for existing deployments.

Follows #969 (PR A: runtime-kind schema + launch seam). Reviewed plan in #958.

## What this does

**Discriminated union + schema v24** (`d01f41b`)
- `AgentRuntimeSpec` is now a real `"tmux" | "herdr"` discriminated union. tmux variant keeps its three locators required; herdr variant carries `herdrSession` / `herdrConfigDir` / `herdrPaneId` / `herdrShellPid`. `terminal_id` is deliberately never persisted — it changes on every Herdr server restart (spike finding, #958).
- Schema v23 → v24: table rebuild widens `CHECK(runtime_kind IN ('tmux','herdr'))`, relaxes tmux columns to nullable, adds nullable herdr columns, and adds per-kind payload consistency CHECKs so a row can never load as a spec the types can't represent. This is the rebuild PR A deliberately deferred (review decision on #969). Persist/read validate per kind.

**Herdr worker runtime controller** (`bb3299a`)
- Implements the #969 seam: ensure the pinet-owned Herdr server session (dedicated `XDG_CONFIG_HOME` with `pane_history` enabled), create the worker workspace with `--env` at pane creation (env cannot be injected later — `pane run` only types into the shell), capture pane id + shell PID into the spec, run the launcher.
- Cleanup fence: probe the pane, verify the observed shell PID matches the recorded `herdrShellPid`, refuse loudly on mismatch, and only then `pane close`. Missing pane/session classified as already-cleaned (herdr analogue of the tmux `can't find session:` handling); other failures propagate.
- Injected command runner, mirroring the tmux controller's testability pattern.

**Config-driven selection, fail-not-substitute** (`89f9aa4`)
- New setting `subtreeWorkerRuntime?: "tmux" | "herdr"`, **default `"tmux"`** — off by default per #958. Read once at spawn; everything downstream (probe/cleanup/teardown/persist) dispatches on the RECORDED kind, never re-reads config.
- An explicitly configured backend that cannot initialize fails with a clear error; it never silently substitutes tmux (#958 safety rule).

**Runtime-aware touchpoints** (the PR-B checklist from #969)
- `a773317` — broker-stop teardown enumerates recorded runtime specs and dispatches each child to its own controller (was: `childTmuxSessions()` + hardcoded tmux cleanup).
- `b71ce98` + `868d616` — `PINET_LAUNCH_SOURCE` gains herdr values; subtree herdr children use `subtree-broker-herdr`, mirroring `subtree-broker-tmux`; all broker-managed checks accept both families.
- `37cb6ee` — spawn results/timeout errors/user-facing messages carry the recorded runtime kind and the controller's monitor hint (`herdr session attach …` for herdr) instead of hardcoded tmux wording.
- `be8d0fd` — neutral `runtimeKind`/`runtimeLocator` across session search, hibernate eligibility, session formatting, and broker registration. Legacy `tmuxSession` is retained on the wire for compatibility, back-filled for old clients, and emitted only for tmux-kind sessions.
- `868d616` — `sessions` search tool accepts `runtime_locator` alongside the legacy `tmux_session` filter; session lines display the recorded kind's locator (tmux output unchanged).

**Mixed-fleet coexistence test** (`db3df8b`) — the one #958 explicitly requires: one tmux worker and one herdr worker under the same subtree broker; asserts per-kind persisted specs, teardown dispatching each child to its own controller via the fake runners' command logs, and correct kinds/hints in spawn results.

## Checklist disposition (from #969)

| Touchpoint | Status |
|---|---|
| Broker-stop teardown enumeration | Done (`a773317`) |
| `pinet-tools` tmux hint strings | Done (`37cb6ee`, `868d616`) |
| `runtime-mode` / `runtime-agent-context` launch source | Done (`b71ce98`, `868d616`) |
| `socket-server` registration params | Done, wire-compatible (`be8d0fd`) |
| `pinet-session-formatting` | Done (`be8d0fd`) |
| `broker-core/lifecycle` locator eligibility | Done (`be8d0fd`) |
| `prompts/broker/tmux.md` + prompt loader | Out of scope: central-broker launch flow, not the subtree spawn seam; still tmux-only |

## Out of scope / known limits

- **Hibernation on Herdr** is #962. Herdr specs hit an explicit "hibernation unsupported on this runtime" boundary rather than a crash or silent skip.
- The registration follow-nudge (`send-keys` retry) is tmux-only; herdr children rely on the launcher-run `/pinet follow`. A `pane run` nudge is a possible follow-up if registration timeouts show up in practice.
- Default stays tmux. The flip is a one-line config change once this has real usage (#958).

## Verification

- Full prepush (lint, typecheck, complete test suite) green on the final push — no `--no-verify`.
- Focused: schema/migration round-trips (v23→v24 with existing tmux rows, per-kind persist/read, corrupt-row rejection), herdr controller launch/cleanup/fence/absence classification, config selection + fail-not-substitute, teardown dispatch, mixed-fleet coexistence.


tmustier:
# PR #972 pre-merge review

Reviewing `feat/958-herdr-backend` against `9521836` (read-only).

## Findings (incremental)

### BLOCKING — A Herdr child can inherit the parent tmux locator and later be dispatched to tmux cleanup

`slack-bridge/subtree-broker-runtime.ts:478-486,551-560,743-759,1005-1040` / `slack-bridge/runtime-agent-context.ts:404-415` / `slack-bridge/broker/socket-server.ts:791-808`

The dedicated Herdr server is spawned with the parent process environment, and pane creation only overlays `childLaunchEnv`; it does not clear `PINET_TMUX_SESSION`. A subtree parent normally launched by tmux therefore passes its own tmux session into the Herdr child. `getAgentMetadata()` records that inherited value as `tmuxSession`, and registration compatibility back-fill can classify it as tmux. If the in-memory `spawnedWorkers` record is unavailable and no durable spec exists (durable persistence is activation-gated/default-off), `childRuntimeSpecs()` falls back solely from metadata `tmuxSession` and constructs a tmux spec. Broker-stop can consequently send the Herdr child to the tmux controller and kill the parent’s tmux session. This violates the recorded-kind dispatch invariant. I reproduced the inheritance against installed Herdr 0.7.5: a server started with `PINET_TMUX_SESSION=parent-stale`, then `workspace create --env PINET_LAUNCH_SOURCE=subtree-broker-herdr`, yielded both variables in the pane. Clear/sanitize the tmux-only variable for Herdr panes (and ensure restart fallback cannot infer tmux when recorded `runtimeKind` is Herdr).

### SHOULD-FIX — Read validation accepts empty runtime locators that persist validation rejects

`broker-core/schema.ts:2721-2741,2851-2883`

`upsertAgentRuntimeSpec()` rejects empty tmux/Herdr locator strings, but `getAgentRuntimeSpec()` checks those columns only for `NULL` (plus Herdr PID validity); both paths also accept whitespace-only locators. A v23 row with an empty tmux locator, or a constraint-bypassed/corrupt Herdr row with an empty locator, therefore loads as a supposedly valid discriminated spec. Make read and persist validation consistently require non-blank per-kind strings. The new corruption test only covers non-null opposite-kind data and does not exercise malformed values within the selected kind.

### SHOULD-FIX — Any invalid runtime setting selects Herdr

`slack-bridge/subtree-broker-runtime.ts:1271-1277` / `slack-bridge/helpers.ts:146-152`

Settings are parsed from JSON via a type assertion, so the TypeScript union is not runtime validation. After the nullish default, the selection ternary treats every configured value other than exact `"tmux"` (for example a typo such as `"tmuux"`, a number, or an object) as Herdr. That undermines the default-off/explicit-opt-in safety claim. Validate the value and either default only when absent or fail clearly for an unsupported configured value.

### SHOULD-FIX — Herdr launch failures before PID capture return unusable cleanup handles

`slack-bridge/subtree-broker-runtime.ts:516-583,592-597,1328-1334` / `slack-bridge/pinet-tools.ts:432-439,2927-2937`

Every launch failure is wrapped with `state=launched_unregistered`, exposed with a retry handle, and classified with an instruction to use that handle. But before a valid pane generation is captured—including the explicitly tested “Herdr unavailable” case—the worker record has a null pane and/or PID. The retry path calls `cleanup()`, which always rejects such a record, so the advertised retry cannot relaunch. In the post-create case (`process-info` failure, malformed JSON, or temporarily missing PID), it also leaves a workspace that this path can never clean. The controller test covers only a failure after both identifiers were captured. Distinguish definitely-not-launched failures from ambiguous ones, and add bounded PID recovery or safe rollback plus retry-path tests for failures before and after workspace creation.

## Validated areas

- Schema v24 rebuild copies every v23 column, recreates the primary key, and its SQL CHECK correctly requires all selected-kind columns while forcing all opposite-kind columns to `NULL`; the v23→v24 tmux migration is transactional.
- Installed Herdr 0.7.5 emits the exact missing-session and `pane_not_found` signatures recognized here. The classifiers are narrow enough that ordinary transport errors are not suppressed. Refusing cleanup on shell-PID mismatch is the correct fail-safe: mismatch means the pane generation is no longer the recorded owned generation, so treating it as absent could close an unrelated shell.
- Runtime selection is read once per new spawn. Record-before-launch, fence-before-retry-cleanup, and per-kind dispatch through the in-memory/durable spec are otherwise preserved; `subtreeWorkerRuntime` has no downstream re-read.
- Legacy registration with only `tmuxSession` is back-filled; search emits legacy `tmuxSession` only for tmux. No consumer matches the old `missing_tmuxSession` reason string.
- The nine commits are coherent and the diff contains no stray files or whitespace errors.

## Verification

- `pnpm exec vitest run broker-core/hibernation-schema.test.ts slack-bridge/subtree-broker-runtime-herdr.test.ts slack-bridge/subtree-broker-runtime.test.ts slack-bridge/broker/integration.test.ts` — 133 tests passed.
- `pnpm --filter @pinet/broker-core typecheck` — passed.
- `pnpm --filter @pinet/slack-bridge typecheck` — passed.
- `git diff --check 9521836...HEAD` — passed.
- Live Herdr 0.7.5 probes verified absence signatures and reproduced inherited `PINET_TMUX_SESSION`; the temporary session/config were stopped and removed.

## Overall verdict: REQUEST-CHANGES

The cross-kind cleanup path can kill the wrong tmux session and is blocking. The three SHOULD-FIX issues are bounded but should be addressed alongside it while the backend/error contract is being introduced.

tmustier:
Follow-up review of the fix commits (868d616..a90bbff) against the REQUEST-CHANGES review above. All four original findings verified FIXED (detail below, including a live Herdr 0.7.5 probe of the env override + launcher unset). The re-review surfaced one further leak in the same family — the shared Herdr server also passes stale `PINET_BROKER_AGENT_ID`/`PINET_LANE_ID` into later panes — now fixed in 93ac31b: every per-launch Pinet identity var not explicitly set by the child launch env is cleared at both pane creation (`--env KEY=`) and in the launcher script (`unset`), with assertions in the herdr controller and spawn regression tests. Full prepush suite green.

# PR #972 fix-commit re-review

Reviewing fix range `868d616..a90bbff` read-only against the original REQUEST-CHANGES review.

Original review read: https://github.com/gugu91/pinet/pull/972#issuecomment-5123884587

## Findings (incremental)

### 1. BLOCKER — FIXED

All three defenses are present:

- Herdr workspace creation passes `--env PINET_TMUX_SESSION=` after the child env, and the Herdr launcher explicitly runs `unset PINET_TMUX_SESSION` before `exec pi` (`slack-bridge/subtree-broker-runtime.ts:552-563,713-722`). Thus even though the dedicated Herdr server command itself inherits the parent environment, the pane gets an empty override and the launched agent gets no variable. `runtime-agent-context` also trims empty strings to `undefined`, so an empty value cannot become either `tmuxSession` or `runtimeLocator` (`slack-bridge/runtime-agent-context.ts:407-416`).
- Herdr launch sources are classified as `runtimeKind: "herdr"` and never emit tmux metadata, even if a stale parent tmux value exists (`slack-bridge/runtime-agent-context.ts:407-416`). The tmux branch still records the prior `tmuxSession` value, now also making its kind/locator explicit.
- Restart metadata fallback skips tmux inference when either `runtimeKind` says Herdr or `launchSource` alone is `broker-herdr`/`subtree-broker-herdr` (`slack-bridge/subtree-broker-runtime.ts:1033-1083`). The stop regression test keeps the parent session live and asserts no tmux kill (`slack-bridge/subtree-broker-runtime.test.ts:666-699`).

Runtime-locator leak audit: `PINET_SOCKET_PATH`, parent/root/spawner IDs, launch source, launch ID, and role are child-overridden; an explicit lane is also overridden. No other inherited variable can be interpreted as a tmux locator, and no Herdr path legitimately consumes `PINET_TMUX_SESSION`. A separate non-locator identity leak is reported below.

### 2. Blank runtime locators — FIXED

Both persistence and read paths require all three selected-kind locator strings to remain non-blank after `trim()`, for tmux and Herdr (`broker-core/schema.ts:2721-2745,2855-2902`). Corruption tests cover each within-kind string locator plus invalid Herdr PIDs; persist tests cover whitespace for each string field (`broker-core/hibernation-schema.test.ts:349-397`).

### 3. Runtime setting validation — FIXED

Only an absent value defaults to tmux; exact `"tmux"` and `"herdr"` select those controllers; every other JSON-representable setting value throws an error containing the rendered value and both valid options before either controller runs (`slack-bridge/subtree-broker-runtime.ts:1315-1327`). Existing/default and explicit tmux/Herdr tests cover the three valid cases; the fix adds invalid string, number, and object coverage (`slack-bridge/subtree-broker-runtime.test.ts:255-335`, plus default tmux spawn coverage earlier in the suite).

### 4. Incomplete Herdr launch rollback — FIXED for the reported failure timings

A pre-workspace failure leaves both generation fields null and cleanup is now a successful no-op. Once a valid pane ID is returned, it is assigned before PID probing; any process-info throw, malformed process-info JSON, or absent/invalid PID attempts pane close and clears `herdrPaneId`. A non-absence rollback error preserves the pane ID and includes it in the loud launch error (`slack-bridge/subtree-broker-runtime.ts:552-608,617-623`). Retry tests exercise failure before create and failure after create/before PID, and controller tests cover rollback success, rollback failure, and no-generation cleanup (`slack-bridge/subtree-broker-runtime.test.ts:358-471`; `slack-bridge/subtree-broker-runtime-herdr.test.ts:117-170`). The assignment order has no in-memory half-generation state except the deliberate failed-rollback state.

Residual edge: malformed/missing `workspace create` JSON after Herdr has in fact created a pane remains intrinsically unlocatable because the pane ID was never returned. This was not made worse by the fix and is outside the safely rollbackable valid-ID path.

### 5. Mechanical commits — FIXED / no behavioral change

`d0c758a` adds the established lint waiver and narrows a compile-time settings assertion; the runtime value and validation are unchanged. `a90bbff` changes test fixture typing only.

### 6. Fix regression risk — TMUX path preserved; Herdr tmux override safe

Tmux-managed workers still trim and record `tmuxSession` exactly as before, now with explicit `runtimeKind: "tmux"` and matching `runtimeLocator`; the regression test asserts all three (`slack-bridge/runtime-agent-context.ts:407-416`; `slack-bridge/runtime-agent-context.test.ts:658-695`). Herdr launch sources deliberately ignore even a non-empty inherited tmux session. No Herdr runtime path reads `PINET_TMUX_SESSION`; live Herdr 0.7.5 probing confirmed `--env PINET_TMUX_SESSION=` makes the pane value empty and the generated launcher's `unset` removes the variable entirely before `pi`.

## New finding

### SHOULD-FIX — The Herdr server can still leak other per-agent identity into panes

`slack-bridge/subtree-broker-runtime.ts:291-314,479-488,552-563,686-723` / `slack-bridge/runtime-agent-context.ts:405-435`

The owned Herdr server is still spawned with all of `process.env`. Pane creation overrides most child identity, but not `PINET_BROKER_AGENT_ID`; it also omits `PINET_LANE_ID` when the spawn input has no lane. The launcher does not clear either. A long-lived server first started by a broker-managed/lane-bound parent can therefore stamp later Herdr children with that server starter's stale `brokerManagedBy` and lane. `getAgentMetadata()` records both values. This is especially nondeterministic because the shared `pinet-workers` server can outlive and serve other subtree parents. Explicitly set the child broker owner and clear optional per-launch identity when absent (or sanitize all non-inherited `PINET_*` identity at the pane/launcher boundary), with a server-inheritance regression test.

This does not reopen the original cross-kind tmux cleanup bug—the runtime-kind and launch-source defenses prevent that—but it is a remaining identity leak in the exact inherited-server path requested for audit.

## Verification

- `pnpm exec vitest run --config vitest.config.ts broker-core/hibernation-schema.test.ts slack-bridge/subtree-broker-runtime.test.ts slack-bridge/subtree-broker-runtime-herdr.test.ts slack-bridge/runtime-agent-context.test.ts` — 87 passed.
- `pnpm --filter @pinet/broker-core typecheck` — passed.
- `pnpm --filter @pinet/slack-bridge typecheck` — passed.
- `git diff --check 868d616...a90bbff` — passed.
- Live isolated Herdr 0.7.5 probes: inherited `PINET_TMUX_SESSION=parent-stale` became empty via the workspace override, then absent after the launcher-equivalent `unset`. A second probe reproduced the remaining leak: unoverridden `PINET_BROKER_AGENT_ID=server-starter` and `PINET_LANE_ID=server-lane` both appeared in the pane. Temporary sessions/configs were removed.

## Overall verdict: REQUEST-CHANGES

All four original review findings are closed, and the two follow-up commits are mechanical. However, the remaining Herdr-server inheritance of `PINET_BROKER_AGENT_ID` and optional `PINET_LANE_ID` should be fixed before approval because it records stale per-agent ownership/routing identity on later children.

PR #1006, opened 2026-08-23
fix(model-aware-compaction): avoid native compaction race

### LINKED ISSUE #1005: fix(model-aware-compaction): avoid duplicate compaction after native auto-compaction
## Maintainer approval

Approved in the active maintainer session.

## Problem

The model-aware compaction extension triggers from `agent_end`, but Pi may still auto-compact and retry after that event. When native compaction wins the race, the extension calls `ctx.compact()` against a branch whose latest entry is already a compaction, producing duplicate UI errors:

```text
Compaction failed: Already compacted
Model-aware compaction failed: Already compacted
```

This occurs frequently where native `reserveTokens` and model-aware thresholds overlap.

## Proposed fix

- evaluate proactive compaction from `agent_settled`, after Pi automatic continuation/compaction has settled
- treat `Already compacted` as an idempotent outcome rather than a retryable failure
- add regression tests for event timing and duplicate-compaction handling
- update lifecycle documentation

## Acceptance criteria

- model-aware compaction does not request compaction from `agent_end`
- eligible settled runs still compact once
- `Already compacted` does not emit the extension error or re-arm an immediate duplicate
- package and repository checks pass

gugu91:
## Summary

- trigger proactive model-aware compaction from `agent_settled` instead of `agent_end`
- skip branches already ending in a compaction before invoking Pi
- treat an `Already compacted` callback as an idempotent outcome without re-arming or duplicate extension errors
- document the corrected lifecycle and add regression coverage

## Root cause

Pi may still auto-compact and retry after `agent_end`. The extension requested a second manual compaction before that automatic lifecycle settled, so Pi rejected it as `Already compacted` and both Pi and the extension displayed errors.

## Validation

- 14 model-aware-compaction tests
- targeted lint and typecheck
- agent-standards lint
- full `pnpm prepush`
- `git diff --check`

Closes #1005

### DIFF
--- a/model-aware-compaction/README.md
+++ b/model-aware-compaction/README.md
@@ -37,7 +37,7 @@ Rules are evaluated in order. `*` wildcards are supported, such as `example-prox
 
 ## Behavior
 
-After each `agent_end`, the extension reads `ctx.getContextUsage()` and the active `ctx.model`. Waiting for `agent_end` is important: tool-using runs can emit several `turn_end` events, and Pi's `ctx.compact()` aborts an active agent operation before compacting. Triggering only after the complete model/tool loop has settled prevents compaction from discarding pending tool work. When usage first exceeds the matching rule's `activeContextTokens`, the extension calls `ctx.compact()`. It prevents duplicate calls while compaction is in flight and re-arms after usage drops below the threshold, the model changes, a session starts, or compaction fails.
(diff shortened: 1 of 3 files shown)


will-jack-and-jilll (approved):
*Will's reviewer agent Race Ferret*

## Verdict
- **Approve**
- Moving the trigger to `agent_settled` matches Pi's actual compaction/retry lifecycle, and the new branch-tail / `Already compacted` guards keep the proactive path idempotent without suppressing normal re-arming. Targeted tests, lint, typecheck, and agent-standards checks all passed.

## Top Notes
- `model-aware-compaction/index.ts:42-92` now waits for `agent_settled`, which avoids the `agent_end` vs native auto-compaction race called out in #1005.
- `model-aware-compaction/index.ts:61-63` mirrors Pi core's own `Already compacted` condition by short-circuiting when the branch already ends in a compaction entry.
- `model-aware-compaction/index.ts:83-90` treats the core `Already compacted` error as idempotent while still re-arming on real failures.
- `model-aware-compaction/index.test.ts:66-191` and `model-aware-compaction/README.md:40-46` cover the lifecycle shift, idempotency, re-arming behavior, and docs update.

Issue #1009, opened 2026-08-25
RFC: integrate session goals with Pinet lanes, RALPH, and lifecycle control


gugu91:
## Status

Parked design RFC. Dynamic goal-budget editing will be implemented first in a separate issue.

## Principle

Pinet may observe, reserve, wake, and report a single-session goal, but must not absorb it into a lane or turn it into multi-agent orchestration.

## Ownership

- `agent-goal` owns objective, budget, usage, evaluation, continuation claims, and completion/blocking decisions.
- Pinet owns lanes, task assignments, agent registration, routing, hibernation, and fenced same-session wake.
- RALPH observes, reports, nudges, and recovers stale continuation lifecycle; it never completes, clears, rewrites, migrates, or decomposes a goal.

## Relationship model

- Pi session: zero or one active goal.
- Lane: multiple participants; each participant may reference its own session goal.
- Task assignment: may seed a goal in the target session and retain only a goal reference.
- RALPH: correlates lane, assignment, workload, and read-only goal telemetry.
- Wake: must return to the same stable agent/session identity.

## Availability semantics

- no goal / complete / cleared: available
- active: reserved and working
- paused: reserved but paused
- blocked or budget-limited: needs attention
- `/pinet free` must refuse while an unfinished goal reserves the session

## Phases

1. Define the integration contract and mismatch taxonomy.
2. Publish read-only goal snapshots to Pinet session/agent metadata and dashboards.
3. Make automatic assignment and `/pinet free` goal-aware.
4. Link lane participants and task assignments by `goalId` + stable session scope.
5. Add a broker-backed continuation/wake adapter with standalone fallback.
6. Add RALPH correlation, stale-claim recovery nudges, and mismatch anomalies.
7. Add audited/versioned Pinet controls; keep Slack read-only until authorization and confirmation semantics are proven.
8. Handle subtree child goals through parent aggregation, not central child ownership.

## Required anomalies

- goal complete while PR/task unfinished
- PR/task complete while goal active
- lane done/cancelled while goal active
- active goal while worker is advertised free
- expired continuation claim
- blocked or budget-limited goal awaiting operator action
- stale lane goal reference

## Rejected designs

- goal as lane state
- RALPH as goal evaluator
- broker decomposition of goals into workers
- goal state copied into lane metadata
- continuation by a different capable worker

## Acceptance constraint

[...]

PR #1011, opened 2026-08-25
feat(agent-goal): allow bounded budget updates

### LINKED ISSUE #1010: feat(agent-goal): allow operators and agents to update turn and token budgets
## Maintainer approval

Approved in the active maintainer session. This is the first implementation step before RFC #1009.

## Problem

A goal budget is fixed at creation. Operators and the goal-bearing agent need to adjust the remaining turn/iteration and token ceilings without clearing and recreating the durable goal.

## Scope

- add atomic, optimistic budget updates to `GoalRuntime`
- expose Pi-local `/goal budget` controls
- expose a model-visible tool so the same agent can request budget changes
- constrain all increases to configured hard ceilings
- allow decreases only when they do not fall below already-accounted usage
- reactivate a `budget_limited` goal when the new budget provides capacity
- preserve standalone operation and storage portability
- emit an auditable structured budget-change event
- update dashboard/modal details and documentation

## Proposed commands

- `/goal budget turns=<n>`
- `/goal budget tokens=<n>`
- `/goal budget turns=<n> tokens=<n>`
- model tool `update_goal_budget` with optional `maxTurns` and `maxTokens`

`turns` maps to the existing settled-run `maxIterations` accounting term; user-facing surfaces should use “turns” consistently while persisted compatibility can retain `maxIterations`.

## Guardrails

- an agent cannot raise a budget above configured environment/settings ceilings
- no implicit unlimited budget
- no budget change may erase usage or bypass atomic settlement accounting
- stale updates must fail and be retried against the current goal version
- changing a budget does not itself claim completion

## Acceptance criteria

- operator and model-tool paths share one runtime method and validation policy
- concurrent budget update/settlement cannot lose token or turn accounting
- increasing an exhausted budget can resume the existing goal safely
- lowering below consumed usage is rejected
- tests cover increases, decreases, hard ceilings, stale CAS, budget-limited recovery, and command/tool parsing
- full prepush and independent review pass

Related design: #1009

gugu91:
## Summary

- add atomic optimistic turn/token budget updates to the standalone goal runtime
- expose operator `/goal budget turns=<n> tokens=<n>` controls
- expose model-visible `update_goal_budget` so the goal-bearing agent can adjust capacity
- preserve pending settlement, terminal-candidate, and continuation accounting across concurrent updates
- reactivate and continue budget-limited goals when new capacity is available
- emit `goal.budget_changed` and update documentation/headless guidance

## Guardrails

- updates remain within configured ceilings
- reductions cannot fall below accounted usage
- complete goals reject budget mutations
- stale CAS updates fail cleanly
- no usage or objective is reset

## Validation

- 65 agent-goal tests
- memory and SQLite atomicity coverage
- package lint/typecheck
- agent-standards lint
- full `pnpm prepush`
- `git diff --check`

Closes #1010
Related: #1009

### DIFF
--- a/agent-goal/README.md
+++ b/agent-goal/README.md
@@ -21,6 +21,7 @@ pi -e ./agent-goal/index.ts
 ```text
 /goal <objective>  Create and immediately start a goal
 /goal              Open the minimal goal window (text in headless modes)
+/goal budget turns=<n> [tokens=<n>]  Change total turn/token ceilings
 /goal pause        Pause automatic evaluation and continuation
 /goal resume       Resume and immediately continue
 /goal complete     Mark complete manually
@@ -33,9 +34,10 @@ Pi's footer is the only persistent goal UI and shows compact status and budget u
 
 Only one goal may exist per Pi session. Clear the existing goal before creating another.
 
-The agent also receives three model-visible tools:
+The agent also receives four model-visible tools:
 
 - `create_goal` — create its own bounded, user-aligned durable goal
+- `update_goal_budget` — change the current goal's total turn or token ceiling
 - `get_goal` — inspect the current objective, status, and budget
 - `update_goal` — optionally attach a `complete` or `blocked` hint for independent verification
 
@@ -51,7 +53,9 @@ PI_AGENT_GOAL_MAX_TOKENS=200000
 PI_AGENT_GOAL_MAX_RUNTIME_MS=14400000
 ```
 
-Iteration and runtime limits are always reliable. Token accounting uses usage reported by Pi providers. The evaluator reviews every settled run, including the final allowed turn, so a completed goal is not incorrectly classified as budget-limited; only another continuation is prevented. The former `PI_AGENT_GOAL_EVALUATION_INTERVAL` setting is accepted for configuration compatibility but no longer changes evaluation frequency.
+Iteration and runtime limits are always reliable. Token accounting uses usage reported by Pi providers. Operators and the goal-bearing agent may update total turn and token ceilings without recreating the goal. Changes are optimistic and atomic, cannot reduce a ceiling below accounted usage, and cannot exceed a configured default ceiling when one exists. Increasing an exhausted budget reactivates the same goal when capacity is available; no budget change alters the objective or erases usage.
+
+The evaluator reviews every settled run, including the final allowed turn, so a completed goal is not incorrectly classified as budget-limited; only another continuation is prevented. The former `PI_AGENT_GOAL_EVALUATION_INTERVAL` setting is accepted for configuration compatibility but no longer changes evaluation frequency.
 
 ## Persistence and recovery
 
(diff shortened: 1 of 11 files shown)


will-jack-and-jilll (changes requested):
*Will's reviewer agent Budget Badger*

## Verdict
- **Block**
- The new budget-mutation path still has two atomicity holes. A concurrent budget change can either drop a just-settled turn entirely or silently kill continuation retries while the goal stays active.

## Top findings
- `agent-goal/runtime.ts:372` — `settle()` returns early when `appendPendingEvaluation()` loses a version race to `updateBudget()`. That drops the settled turn/token delta instead of re-binding it to the new goal version. I reproduced this against both `MemoryGoalStorage` and `SqliteGoalStorage`; the goal budget updated, but `usage` stayed `0/0` after `runtime.settle(...)`.
- `agent-goal/runtime.ts:577` + `agent-goal/memory-storage.ts:201` + `agent-goal/sqlite-storage.ts:739` — an in-flight continuation attempt can overwrite the claim back to the old `goalVersion` after `updateBudget()` has already moved it forward. On the next loop the runtime deletes that stale claim and exits without creating a fresh one, so an active goal stops retrying. I reproduced this on both adapters with `retryPolicy.maxAttempts = 2`; `continueIfIdle()` ran once instead of twice.

Please fix the version-race handling and add regression coverage for both adapters.

will-jack-and-jilll (changes requested):
*Will's reviewer agent Quota Ferret*

Re-review after `5d04607`: the two original version-binding races look fixed, and the new deterministic tests plus my old repro scenarios now pass.

I still found one blocking budget-semantics hole:

- `agent-goal/runtime.ts:191` + `agent-goal/runtime.ts:340` — lowering `maxIterations`/`maxTokens` to the current accounted usage while a turn is already in flight immediately flips the goal to `budget_limited`, and the later `settle()` returns early because the goal is no longer `active`. The just-finished turn's iteration/token usage is therefore dropped instead of being charged.

Why this matters: the README still promises that budget changes do not erase usage and that every settled iteration/token delta is charged. Right now an operator or the goal-bearing agent can lower the budget mid-run and silently undercount the work that just completed.

I reproduced this on both `MemoryGoalStorage` and `SqliteGoalStorage`: after one accounted turn, lowering the budget to that accounted usage and then settling the next run leaves usage stuck at the old values.

Please either:
- reserve the in-flight turn when validating/lowering budgets, or
- allow the already-running turn to settle/account even if the budget change already moved the goal to `budget_limited`, then keep the goal budget-limited for further continuation.

will-jack-and-jilll (approved):
*Will's reviewer agent Throttle Otter*

Re-review after `853d321`: approving.

The last blocker is fixed. `agent-goal/runtime.ts:165-199` now rejects budget reductions on active goals unless they leave room for the current turn, so `updateBudget()` can no longer force an `active -> budget_limited` transition that causes the later `settle()` to skip accounting.

I re-ran the remaining lowering repro on both `MemoryGoalStorage` and `SqliteGoalStorage`: equal-to-accounted reductions are rejected, and the subsequent settled turn is still fully charged. The package suite (`68` tests) and full `pnpm prepush` also pass.

PR #1015, opened 2026-08-25
fix(agent-goal): start idle goals with pending delivery

### LINKED ISSUE #1014: fix(agent-goal): prevent idle goal continuation from stalling on pending messages
## Summary
An operator-created goal can remain at zero accounted turns indefinitely even while its Pi session is idle. The durable continuation claim repeatedly records `The goal session is busy`.

## Observed evidence
- goal status: active
- accounted iterations: 0
- continuation: deferred, attempt 0
- session is reported idle externally
- `last_error`: `The goal session is busy`

## Root cause hypothesis
The default continuation adapter requires both `ctx.isIdle()` and `!ctx.hasPendingMessages()`. Pi can report an idle session with queued/pending delivery, so the claim retries forever and no agent turn begins. This is initiation/idle gating, not settlement turn accounting: the zero count is expected until an agent run settles.

## Acceptance criteria
- Operator-created goals reliably initiate in the same stable session.
- Pending delivery cannot strand an idle session indefinitely.
- Preserve same-session ownership and durable idempotent claims.
- Do not double-trigger or double-charge turns.
- Add deterministic regressions for idle+pending and busy delivery behavior.

Maintainer approval: requested interactively with “are you fixing?”

gugu91:
Closes #1014.

## Summary
- treat Pi `isIdle()` as authoritative for goal initiation
- do not strand operator-created goals on an idle session whose pending-delivery flag remains set
- add a red/green regression that initiates the goal and verifies exactly one settled turn is charged

## Evidence
- Before fix: new regression failed because `sendMessage` was called 0 times.
- After fix: 72/72 agent-goal tests pass and the run settles at exactly one iteration.
- Separate live-panel clean-idle smoke test initiated immediately and executed one model turn.
- Full `pnpm prepush` and `git diff --check` pass.

### DIFF
--- a/agent-goal/index.test.ts
+++ b/agent-goal/index.test.ts
@@ -103,6 +103,57 @@ describe("registerAgentGoal", () => {
     },
   );
 
+  it("starts an operator-created goal when the idle session reports pending delivery", async () => {
+    const handlers = new Map<string, GoalEventHandler>();
+    const commands = new Map<string, RegisteredCommand>();
+    const sendMessage = vi.fn();
+    const pi = {
+      on(name: string, handler: GoalEventHandler) {
+        handlers.set(name, handler);
+      },
+      registerTool: vi.fn(),
+      registerCommand(name: string, command: RegisteredCommand) {
+        commands.set(name, command);
+      },
+      sendMessage,
+    } as object as ExtensionAPI;
+    const context = {
+      hasUI: true,
+      isIdle: () => true,
+      hasPendingMessages: () => true,
+      sessionManager: { getSessionId: () => "session-1" },
+      ui: {
+        setStatus: vi.fn(),
+        setWidget: vi.fn(),
+        notify: vi.fn(),
+      },
+    } as object as ExtensionCommandContext;
+    const storage = new MemoryGoalStorage();
+    registerAgentGoal(pi, {
+      storage,
+      evaluator: {
+        evaluate: vi.fn().mockResolvedValue({ outcome: "complete", reason: "verified" }),
+      },
+    });
+    const command = commands.get("goal");
+    if (!command) throw new Error("goal command was not registered");
+
+    await command.handler("verify operator goal start", context);
+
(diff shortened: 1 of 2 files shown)


will-jack-and-jilll (changes requested):
*Will's reviewer agent Spline Ferret*

Verdict: **Block**

Top findings:
- `agent-goal/index.ts:139` — removing `hasPendingMessages()` is unsafe with Pi's current idle followUp delivery. On an idle session, `sendMessage({ deliverAs: "followUp", triggerTurn: true })` prompts immediately instead of queueing, so any already-queued follow-up can still drain afterward and produce more than one settled turn.
- `agent-goal/index.test.ts:106` — the new regression stubs `sendMessage` and manually simulates one settlement, so it does not prove initiation plus exactly one settled turn under real Pi queue semantics.

I reproduced the queueing behavior against `pi-agent-core`: a queued follow-up plus an idle custom continuation prompt produced two turns in the same session. I’d keep the pending-message guard (or switch to an idle-safe queued delivery path) before merging.

will-jack-and-jilll (approved):
*Will's reviewer agent Queue Opossum*

Verdict: **Approve**

PR reviewed: #1015

Re-reviewing against the actual goal-accounting contract resolves the prior blocker.

- `agent-goal/index.ts:139` is safe under this contract: goal usage is charged once per `agent_settled` run, not per internal prompt.
- `agent-goal/index.test.ts:106` correctly models the extension-observed lifecycle by asserting one settlement and one charge when idle start happens with pending delivery.
- Verification passed:
  - `pnpm --filter @pinet/agent-goal typecheck`
  - `pnpm --filter @pinet/agent-goal test`
  - AgentSession repro: queued follow-up + idle trigger => `promptCalls=2`, `agent_settled=1`

Summary: Pi bundles the queued follow-up into the same session-level settled run, and `agent-goal/runtime.ts:333-360` accounts exactly one iteration for that settlement. No remaining blocker on this PR.

Issue #1031, opened 2026-09-11
RFC: Pinet v2 with channels, Markdown work tracking and managed agent runtimes


gugu91:
## Proposal

Build Pinet v2 around four simple ideas:

- Projects contain tasks.
- Each project has a shared channel.
- Projects and tasks are Markdown documents.
- Pinet starts, tracks, stops and resumes agents.

Agents should organise the work themselves. Pinet should provide the shared records, communication and process management they need. Goals remain separate and unchanged.

This replaces the earlier proposal for strict agent hierarchies, assignment workflows and special PM permissions. It is still a design proposal, not approval to start building.

## Projects and tasks

A project has an ID, one channel and a Markdown document. A task has an ID, belongs to one project and also has a Markdown document.

Tasks do not have subtasks. Agents record owners, status, checklists, dependencies and results in the Markdown itself. Pinet does not need to interpret or enforce these fields.

Agents can link related projects and tasks. For the first version, ordinary reads and writes are enough. We accept that two agents editing the same document at once may overwrite each other's changes.

## Channels

A project's channel is the main place for its agents to talk. Agents can join channels, search their history and contact agents in other channels. An agent starts with one home channel, but that does not restrict who it can talk to.

Channels should support mentions. We still need to decide:

- which messages wake an agent
- how threads work
- how messages between channels appear

We should not wake every agent whenever someone posts in a channel.

## Project managers

A project manager is an ordinary agent appointed by the user. Pinet records who it is so that new human messages reach the right agent.

The PM can organise tasks, ask other agents to work on them and check the results. Any agent can do the same things: there is no special PM runtime or exclusive permission to delegate work.

## Goals

Tasks and goals solve different problems.

A task records work that the team wants to track. A goal keeps one agent working towards an outcome.

Keep the existing goal extension unchanged. A task does not need to become a goal. Spawning an agent does not make you the owner of its goal. Completing a goal does not automatically close a task or stop the agent.

An agent can finish one task and then take another without losing its session context.

## Agent runtimes

Any agent can ask Pinet to start another agent in a channel with a prompt. Pinet carries out the request and tracks the process. It does not decide what work should be delegated.

Pinet should store:

- the agent, its channel and who requested it
- its Pi session and working directory or worktree
- how it was started and the process or pane that belongs to it
- when it started, when it last checked in and its current state

This information belongs in SQLite. Who started an agent is useful history, not a management hierarchy.

[...]

PR #1036, opened 2026-09-14
feat(work): add independent Markdown projects and tasks

gugu91:
## Issue / problem solved

Refs #1031. This is the **Work** layer of native GitHub stack #1040, replacing the combined PR #1035. Base: main.

Add `@pinet/work`: independent Markdown projects/tasks with CRUD, search, native SQLite persistence and a Cloudflare Durable Object adapter. Includes the shared Hono/emulator dependency lock entries and manifest-derived build completeness test.

## Design and scope

This is a history-only split of the reviewed implementation, not a rewrite. Existing extensions remain untouched. New tool surfaces use compact dispatchers with `help`; PM is optional. Package build/workspace wiring is introduced alongside each package so every intermediate checkout works.

Stack order: #1036 Work → #1037 Chat → #1038 Slack → #1039 PM. The final tree at `528e02064674a9e252bcefad46770488effc899d` is byte-for-byte identical to the previously approved `1f10fb86aea41d797058b95170cc180931cad91a`.

## Testing / validation

Run independently at this layer:
- `pnpm install --frozen-lockfile`, including postinstall builds: passed.
- `pnpm prepush` (lint, agent-standards, strict typecheck, workspace/script tests): passed.
- `pnpm format:check`: passed.

The aggregate implementation passed local Miniflare persistence/runtime tests and Cloudflare bundle dry-runs. Independent approval and the original validation record are preserved at https://github.com/gugu91/pinet/pull/1035#issuecomment-5647128682. GitHub CI runs separately on each new PR.

## Review notes

Last-write-wins Markdown, no enforced task status or goal lifecycle. Test local and Durable Object persistence, project deletion cascades, search pagination and token parsing.

No merge, publication, deployment or existing-extension migration is included. No new implementation alternatives were introduced by splitting the history.


work/storage.ts
```diff
@@ -0,0 +1,125 @@
...
+    return (rows as object[]).map((row) => this.task(row)!);
+  }
+  deleteTask(id: string): boolean {
+    return this.database.prepare("DELETE FROM pinet_tasks WHERE id=?").run(id).changes > 0;
+  }
+  search(query: string, limit: number, offset: number): { projects: Project[]; tasks: Task[] } {
+    const rows = this.database
+      .prepare(
+        `
+      SELECT 'project' AS kind,id,markdown,external_channel,NULL AS project_id,created_at,updated_at FROM pinet_projects WHERE instr(lower(markdown),lower(?))>0
```
will-jack-and-jilll: can we separate sql into its own files? and we need prettier on it. also separate migras please

work/server.ts
```diff
@@ -0,0 +1,186 @@
...
+      message.includes("not found") ? 404 : 400,
+    );
+  });
+  app.get("/health", (c) => c.json({ status: "ok" }));
+  app.post("/v1/projects", async (c) => {
+    const body = parse(await c.req.text()),
+      timestamp = now();
+    const value: Project = {
+      id: id(),
+      markdown: required(body.markdown, "markdown"),
```
will-jack-and-jilll: no boundaries? have we thought about security?

work/server.ts
```diff
@@ -0,0 +1,186 @@
+import { timingSafeEqual } from "node:crypto";
+import { Hono } from "hono";
+import type { Context } from "hono";
+import type { Project, Task, WorkStorage } from "./domain.js";
+type JsonValue = string | number | boolean | null | JsonValue[] | { [key: string]: JsonValue };
+type JsonObject = { [key: string]: JsonValue };
+export function parseTokens(source: string): string[] {
```
will-jack-and-jilll, line 19: we need prettier here unlegible

work/index.ts
```diff
@@ -0,0 +1,163 @@
...
+    });
+    const value: { data?: object | null; error?: { message: string } } =
+      response.status === 204
+        ? { data: null }
+        : ((await response.json()) as { data?: object; error?: { message: string } });
+    if (!response.ok)
+      throw new Error(value.error?.message ?? `Work request failed (${response.status})`);
+    return value;
+  }
+  pi.registerTool({
```
will-jack-and-jilll, line 62: wonder if we make this an MCP so we get real inter agent compat? we could still host the MCP server as an extension for pi?

will-jack-and-jilll: but in case of cloudflare deployment the server would be in the worker. I guess we would have a daemon locally anyhow, so we don't really need to do any of that? we host the MCP server on the daemon?

PR #1042, opened 2026-09-14
fix(agent-goal): distinguish lifecycle prompts

### LINKED ISSUE #1041: Make agent-goal lifecycle prompts explicit
## Context

Agent goal lifecycle messages currently make a newly created or edited goal look like an ordinary continuation. This can confuse an agent into continuing superseded work.

## Required behavior

- A newly created goal emits an explicit `[agent-goal.started]` message stating that it is a new goal and supersedes prior work.
- An objective edit emits `[agent-goal.updated]`, clearly invalidating the prior objective and its completion criteria.
- Only subsequent runs of the unchanged goal emit `[agent-goal.continuation]`.
- Keep all three prompts concise and treat the objective as user-provided data.
- Add regression tests for lifecycle selection and wording.

## Acceptance criteria

- Agents can unambiguously distinguish started, updated, and continued goals.
- Started/updated prompts explicitly prevent pursuit or completion of the superseded objective.
- Existing continuation durability and idempotency behavior remains intact.
- Package lint, typecheck, and tests pass.

gugu91:
Closes #1041

## Summary
- label new, updated, and unchanged continuation turns distinctly
- make started/updated prompts explicitly supersede prior work and completion criteria
- persist lifecycle kind in continuation claims across retries and restarts
- start an updated-objective turn immediately through the existing durable continuation path

## Validation
- `pnpm prepush`
- agent-goal: 85/85 tests passed


will-jack-and-jilll (commented):
*Will's reviewer agent Prompt Possum*

## Verdict: **Block**

The prompt wording and SQLite column migration look sound, and package tests/lint/typecheck pass, but lifecycle provenance is not durable enough to meet #1041.

### Must fix
1. **`agent-goal/runtime.ts:303-306, 698-719` — an objective edit racing an in-flight evaluation can lose `updated` and launch a second turn.** `updateDetails()` creates an `updated` claim for version N, but the rebased pending evaluation can commit version N+1; `continueWithClaim()` then deletes the now-stale updated claim and creates a `continuation` claim. A source-level reproduction produced calls `[updated@v2, continuation@v3]`; when the first call was busy, the only started turn was the wrongly labeled continuation. Preserve/rebind the pending lifecycle intent atomically across evaluation commits and do not launch a second claim until the updated turn is acknowledged.
2. **`agent-goal/runtime.ts:715-719, 484-505` — `started`/`updated` intent does not exist durably before a claim.** A goal created by `create_goal` and restarted before settlement has no claim, so `recover()` emits `continuation`; likewise an objective edit while snoozed/paused cannot create an active claim and later resumes as `continuation`. Persist the next lifecycle kind independently of an active delivery claim (or atomically with the goal transition), and consume it only when the corresponding turn starts.

### Coverage needed
Add regressions for create→restart-before-settle, deferred updated claim→concurrent evaluation commit/restart, and paused/snoozed objective edit→resume. Also add an actual pre-`kind` SQLite schema migration test; the current test only round-trips a freshly created new-schema row.

### Evidence
- `pnpm --filter @pinet/agent-goal test` — 85/85 passed
- package lint, typecheck, and changed-file Prettier check passed
- all GitHub checks passed
- manual current-source race reproductions confirmed both misclassifications
- manual legacy SQLite migration check preserved the old row with `kind = continuation`

will-jack-and-jilll (dismissed):
*Will's reviewer agent Durable Ferret*

## Updated verdict: **Approve**

Re-reviewed amended commit `45fbebe5e775806e4e384e8002a42f475721169e`. The two prior lifecycle blockers are resolved: `started`/`updated` intent is now persisted independently of delivery claims, goal mutations and evaluation commits rebase matching claims, and acknowledgment atomically consumes only a matching started lifecycle turn.

No remaining actionable findings.

Evidence:
- concurrency/recovery regression harness: 4/4 passed
- full `@pinet/agent-goal` suite: 90/90 passed
- lint, typecheck, Prettier, and `git diff --check`: passed
- legacy pre-`kind` SQLite migration regression: passed
- all GitHub checks: passed

will-jack-and-jilll (changes requested):
*Will's reviewer agent Tag Tapir*

## Updated verdict: **Block**

### Must fix
- **`agent-goal/index.ts:138-145` — the lifecycle tag is not visible to the agent.** Pi's installed `@earendil-works/pi-coding-agent@0.74.0` `convertToLlm()` converts a custom message to a user message containing only `content`; it drops `customType`. With the amended prompts, `agent-goal.started` and `agent-goal.updated` therefore both reach the model as exactly `Objective (user data): …`, so the model cannot distinguish a new goal from an edited one or know that prior work is superseded. Keep a minimal transition marker/boundary in the content, or change the transport so the lifecycle tag is actually included in LLM context.

### Coverage
The current tests assert the outbound `sendMessage` envelope, but not the LLM-visible message. Add a regression at the conversion/integration boundary proving started, updated, and continuation remain distinguishable after `convertToLlm()`.

### Evidence
A direct harness produced identical converted messages for `agent-goal.started` and `agent-goal.updated`; only `Objective (user data): ship` remained. Package tests (90/90), lint, typecheck, formatting, and diff checks pass, but they do not catch this semantic loss.

will-jack-and-jilll (dismissed):
*Will's reviewer agent Boundary Ferret*

## Verdict: **Approve**

Re-reviewed current head `37a0a7542e4ee67b591a09accd409ced865798ce`. The prior model-boundary blocker is resolved: `started`, `updated`, and `continuation` now place their explicit lifecycle tags in `content`, and the installed `@earendil-works/pi-coding-agent@0.74.0` `convertToLlm()` preserves that content as user-message text. A direct harness produced three distinct model-visible strings.

The new regression genuinely calls the package's exported runtime implementation—there is no Vitest mock or alias for `@earendil-works/pi-coding-agent`—and asserts the exact post-conversion role/content for all three tags. Paired exact `sendMessage` assertions cover the production prompt generation that feeds that boundary.

No actionable findings. Verification: focused conversion test passed; full `@pinet/agent-goal` suite passed (91/91); package lint/typecheck, changed-file Prettier, `git diff --check`, and all GitHub checks passed.

will-jack-and-jilll (changes requested):
*Will's reviewer agent Clear Capybara*

## Verdict: **Block**

The new tool is model-visible, status-agnostic, and clearly instructs the model to act only on explicit user request. Successful clears also use the existing durable delete/UI-refresh path. However, the clear operation is not race-safe and can either leave the requested goal running while falsely reporting it missing, or delete a concurrently created replacement goal.

### Must fix
1. **`agent-goal/runtime.ts:406-414`, `agent-goal/index.ts:447-459`, `agent-goal/domain.ts:125` — clear conflates absence with an optimistic-write conflict and does not fence goal identity.** `GoalRuntime.clear()` reads the goal and then deletes by only `scopeId + version`. If an update/evaluation advances the active goal between those calls, deletion returns `false`; `clear_goal` reports “This session has no goal to clear” even though it remains durable and may continue. In an ABA race where the original is cleared and a replacement is created at the same reset version, the scope/version-only delete can remove the replacement even though it was not the goal observed for this request.
   - **Why it matters:** an explicit destructive request may silently fail, and a different durable objective can be destroyed. Pi executes tools in parallel by default, and SQLite can also be shared across overlapping runtimes, so this is a real lifecycle boundary rather than a theoretical type issue.
   - **Suggested fix:** make deletion identity-fenced (`scopeId + goalId + version`) and return a result that distinguishes `cleared`, verified `missing`, and `conflict/replaced`. Retry a bounded same-goal version conflict if appropriate, but never delete a different replacement without a fresh explicit request. Only return the missing-goal message after verifying absence, and refresh UI on every terminal result.

### Coverage needed
- Add regressions for a version advance between read/delete and for clear→replacement ABA; assert the old request cannot leave work running unnoticed or delete the replacement.
- Expand `agent-goal/index.test.ts:63-135` to pin the explicit-request prompt metadata and cover active/blocked/complete deletion plus all dependent state (pending evaluation, candidate, claim, checkpoints), widget/status cleanup, and SQLite cascade behavior. The current test covers only a blocked in-memory goal with one checkpoint.

### Evidence
- Full `@pinet/agent-goal` suite: 92/92 passed
- Package lint, typecheck, Prettier, `git diff --check`: passed
- All GitHub checks: passed
- Focused source-level harnesses reproduced both `false + goal remains` after a version race and successful deletion of a replacement goal after an ABA race

will-jack-and-jilll (approved):
*Will's reviewer agent Race Ferret*

## Verdict: **Approve**

Re-reviewed current head `32cb7d8a3542ad403af77a2bf2e08d0f9265018b`. The prior `clear_goal` race blocker is resolved.

- `GoalStorage.delete` now returns distinct `deleted` / `missing` / `conflict` outcomes and fences `scopeId + goalId + version`.
- SQLite performs the read, identity/version check, and fenced delete under one `BEGIN IMMEDIATE` transaction. An independent-connection harness confirmed that a version advance returns `conflict`, an ABA replacement survives and returns `conflict`, and true delete/missing cases remain distinct.
- `GoalRuntime.clear` reports conflicts as retryable errors rather than absence. Regression tests cover concurrent updates, ABA replacement protection, and active/blocked/complete goals.
- SQLite cascade coverage verifies cleanup of pending evaluations, terminal candidates, continuation claims, and checkpoints.
- The model-facing explicit-user-request guardrails and successful-clear status/widget cleanup remain pinned by tests.

No actionable findings. Verification passed: focused clear/storage/tool tests; full `@pinet/agent-goal` suite (97/97); package lint/typecheck; Prettier; `git diff --check`; and all GitHub checks.

PR #1060, opened 2026-09-21
feat(compaction): separate summary output budget

tmustier:
## Summary

- add optional `model-aware-compaction.summaryReserveTokens`
- use it for selected-model preflight and history/split-turn output budgets
- preserve Pi’s effective reserve when the setting is omitted
- fail selected-model compaction closed for an explicitly invalid value

## Why

Pi 0.86 per-model reserves can express an early native trigger for 1M-context conversation models only with a large `reserveTokens`. Reusing that value for a 128k-context dedicated summarizer makes every request fail preflight. This keeps trigger policy and summarizer output policy independent.

## Verification

- `pnpm check`
- `pnpm test`
- model-aware compaction: 70 tests passed
- pre-push checks passed


tmustier:
Post-merge acceptance on Pi 0.86.1: a forced native mid-turn threshold used the dedicated `summaryReserveTokens` budget, persisted a Gemini-generated compaction, and resumed the active GPT-5.6-sol request. A second real run verified unavailable-primary fallback to GPT-5.6-sol and continuation. The 1M-context trigger-reserve decoupling remains covered by the regression suite.

